PR #9, opened 2022-04-27
Add dark mode

alse:
- Uses code from the official remix dark mode example: https://github.com/remix-run/remix/tree/main/examples/dark-mode
- Use can toggle dark mode, and choice is persisted in cookies
- Dark mode defaults to system setting, and updates when system setting changes
- Flicker free on initial load and subsequent SSR requests

### DIFF
--- a/apps/flow-docs/app/routes/action/set-theme.tsx
+++ b/apps/flow-docs/app/routes/action/set-theme.tsx
@@ -1,0 +1,26 @@
+import type { ActionFunction, LoaderFunction } from "@remix-run/node";
+import { json, redirect } from "@remix-run/node";
+import { getThemeSession } from "~/utils/Theme.server";
+import { isTheme } from "~/utils/ThemeProvider";
+
+export const action: ActionFunction = async ({ request }) => {
+  const themeSession = await getThemeSession(request);
+  const requestText = await request.text();
+  const form = new URLSearchParams(requestText);
+  const theme = form.get("theme");
+
+  if (!isTheme(theme)) {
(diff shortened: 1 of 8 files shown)


apps/flow-docs/app/routes/action/set-theme.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const themeSession = await getThemeSession(request);
+  const requestText = await request.text();
+  const form = new URLSearchParams(requestText);
+  const theme = form.get("theme");
+
+  if (!isTheme(theme)) {
+    return json({
+      success: false,
+      message: `theme value of ${theme} is not a valid theme`,
+    });
```
10thfloor, line 16: Do we need to return a default theme here? Or, is it handled elsewhere?

apps/flow-docs/app/utils/ThemeProvider.tsx
```diff
@@ -0,0 +1,282 @@
...
+  }, []);
+
+  return (
+    <ThemeContext.Provider value={[theme, setTheme]}>
+      {children}
+    </ThemeContext.Provider>
+  );
+}
+
+const clientThemeCode = `
```
10thfloor, line 95: Nice.

10thfloor (approved):

apps/flow-docs/app/routes/action/set-theme.tsx
alse: Yes, the frontend defaults to a light theme

Issue #19, opened 2022-05-11
[Components] Footer


srinjoyc:
# Context

Create the footer component in the "Global Navigation" screen [here](https://www.figma.com/file/aoIxdVZVGltGy4lWetuWKU/Component-Library-(External)?node-id=0%3A1).

# Acceptance Criteria 
- [ ] Desktop and mobile version of the footer should expand into a max of 5 columns on the desktop and overflow onto the next row after that. For mobile, 2 columns before the next row. The first column shows the max number of support links per column.

PR #34, opened 2022-05-16
Neilzon/footer

neilZon:
#19 

Add footer component as outlined in #19 

Note: There were `ul` and `li` styles in `tailwind.css` that I couldn't prevent from being applied unless I removed them from the file, was wondering if they were important?


libs/design-system/src/lib/Components/Footer/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+              <DiscordIcon fill="white" height={28} width={28} />
+            </a>
+            <a href="#" className="px-3">
+              <ForumIcon fill="white" height={24} width={24} />
+            </a>
+            <a href="#" className="px-3">
+              <OnFlowIcon height={28} width={28} />
+            </a>
+            <a href="#" className="pl-3">
+              <Link to="#">Eng</Link>
```
alse, line 139: No need for the <Link> here

libs/design-system/src/lib/Components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { Footer } from '.';
+
+export default {
+  component: Footer,
+  title: 'Components/Footer',
+} as Meta;
+
+const Template = () => (
+  <MemoryRouter>
```
alse, line 12: The MemoryRouter can be removed because we'll be using plain a tags

libs/design-system/src/lib/Components/Footer/index.tsx
```diff
@@ -0,0 +1,170 @@
+import { Link } from 'react-router-dom';
+
+import { ReactComponent as FlowDocsLogo } from '../../Images/logos/flow-docs-logo-light.svg';
+import { ReactComponent as GithubIcon } from '../../Images/social/github-light.svg';
+import { ReactComponent as DiscordIcon } from '../../Images/social/discord-light.svg';
+import { ReactComponent as ForumIcon } from '../../Images/social/forum-light.svg';
+import { ReactComponent as OnFlowIcon } from '../../Images/logos/flow-icon-bw-light.svg';
+
+// reduce repetition of the section layout in Footer component
```
alse, line 9: 👍

libs/design-system/src/lib/Components/Footer/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+export function Footer() {
+  return (
+    <footer className="bg-black px-6 text-white">
+      <div className="container mx-auto">
+        <div className="block px-2 pt-8 pb-6 md:flex md:items-center md:justify-between md:px-4 md:pt-16">
+          <div>
+            <FlowDocsLogo className="h-10 w-min object-contain" />
+          </div>
+          <div className="flex items-center justify-start pt-8 lg:justify-center lg:pt-0">
```
alse, line 125: The `lg:pt-0` should be `md:pt-0` for better vertical alignment at medium widths.

Removing `px-3` on the links below, and adding a gap-6 on this div would also fix the icon left alignment on mobile

alse (approved):
@neilZon looks great, just left a few comments!

libs/design-system/styles/tailwind.css
```diff
@@ -6,19 +6,10 @@
 @import './mdx';
 
 /* MDX styles. TODO: move to ./mdx.css and nest under .mdx-content */
```
alse, line 8: @neilZon could you comment these out instead? They'll have to be moved out to mdx.css and nested

libs/design-system/src/lib/Components/Footer/index.tsx
neilZon: I think I might need some clarification for this comment, not quite sure what its referring to :)

alse: @neilZon hah the github markdown parsed out the `<Link>` from my comment - updated :)

PR #53, opened 2022-05-26
Headers for homepage and landing page

### LINKED ISSUE #43: "Headers" for homepage and landing page

joekrill:
# Adds the header components for the Landing and Homepage and a Button Component

## Button

Adds a shared Button component with `primary` and `secondary` variants:
 
Screen Shot 2022-05-26 at 6 19 51 AM

Screen Shot 2022-05-26 at 6 19 58 AM

## Landing 

Screen Shot 2022-05-26 at 6 01 51 AM

I'm not sure if this is the preferred way we want to handle these background gradients. They were already included in the `/images` folder as SVGs, and using them as CSS `background-image`s seemed like the simplest approach. But I also notice that elsewhere we're using `linear-gradient` rules instead of the images directly. Are the images themselves included just for reference and the preferred way is using a `linear-gradient` rule?

## Homepage

Screen Shot 2022-05-25 at 6 09 36 PM

I took some liberties on the design here a little bit because there were some overflow problems that occur in the mobile view. The designs seem to call for the Title/H1 font to be used here. But in our default mobile viewport size the header text overflows, so we end up with something like this:

Screen Shot 2022-05-25 at 5 44 08 PM

So I reduced the font-size on smaller widths, which isn't included in the design, but allows for the current title to fit: 

Screen Shot 2022-05-25 at 6 09 18 PM

Do we prefer to just use the original design and let the text overflow? I suppose changing the title would work, but with any i18n we'll likely still end up with overflows no matter what. 

---

Closes #43


libs/design-system/src/lib/Components/Button/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    children: React.ReactNode;
+  };
+
+export function ButtonLink({
+  className,
+  variant = 'primary',
+  ...props
+}: ButtonLinkProps) {
+  return (
+    // eslint-disable-next-line jsx-a11y/anchor-has-content
```
alse, line 57: Adding children to the props, and passing children to the `a` tag should satisfy jsx-a11y/anchor-has-content

libs/design-system/src/lib/Components/Button/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+    />
+  );
+}
+
+export type ButtonLinkProps = React.ComponentPropsWithoutRef<'a'> &
+  ButtonBaseProps & {
+    children: React.ReactNode;
+  };
+
+export function ButtonLink({
```
alse, line 51: Nice, good idea splitting this out and having both Button and ButtonLink components

libs/design-system/src/lib/Components/LandingHeader/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  buttonText,
+  buttonUrl,
+  callout,
+  description,
+  gradient,
+  title,
+}: LandingHeaderProps) {
+  return (
+    <LandingHeaderContainer gradient={gradient}>
+      <div className="flex max-w-6xl flex-col justify-items-stretch px-4 py-12 md:px-20 md:py-20">
```
alse, line 23: I just pushed a container tailwind config to the branch. Removing the max-w-6xl and wrapping this with a `<div className="container"` will center everything and make the max width 1140px

libs/design-system/src/lib/Components/LandingHeader/LandingHeaderContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+  };
+
+export function LandingHeaderContainer({
+  className,
+  gradient,
+  ...props
+}: LandingHeaderContainerProps) {
+  return (
+    <header
+      className={clsx(`bg-gradient-${gradient} bg-cover`, className)}
```
alse, line 24: Nice, I think background image is the way to go for these. I think the headers could have more space at the top, and the gradients could move up higher to match the pages in the Flow Dev Portal figma

libs/design-system/src/lib/Components/LandingHeader/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <h1 className="text-h1 mb-6 max-w-full overflow-hidden text-ellipsis !text-4xl md:!text-7xl">
```
alse, line 24: Good call on adjusting these font sizes. I think we can just update the text-h1, we'll probably run into those overflow issues in other places. The mb can be 12 here

libs/design-system/src/lib/Components/LandingHeader/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+          {title}
+        </h1>
+        <div className="flex flex-row items-stretch rounded-lg bg-primary-gray-100/30">
+          <div className="px-5 py-3 md:basis-1/2 md:px-20 md:py-12">
+            <h2 className="text-h2 mb-2 line-clamp-1 md:mb-3 md:line-clamp-2">
```
alse, line 29: Should we remove the line clamp here? Only 1.5 words can be seen on mobile

alse (changes requested):
Looking great @joekrill , just left some small comments

libs/design-system/src/lib/Components/LandingHeader/index.tsx
joekrill: Sure happy to remove that! The reason I added that was because the Figma design had a placeholder title of "Featured callout here two lines." - so I thought that was indicating that this should be limited to 2 lines. And the mobile design uses the title "One liner" so I thought that was indicating just a single line on mobile. But definitely simpler to get rid of the clamp entirely.

libs/design-system/src/lib/Components/Button/index.tsx
joekrill: Ah thanks! I made it a required prop but was still getting the warning. Passing it explicitly works!

alse (approved):
GTM!

PR #68, opened 2022-05-31
Page Content 3 Column Link Card

### LINKED ISSUE #45: "Page Content" 3 Column Link Card - note the hover state

joekrill:
# Adds 3 Column Link Card

Light:

Screen Shot 2022-05-31 at 10 28 10 AM

Dark:

Screen Shot 2022-05-31 at 10 28 16 AM

Mobile:

Screen Shot 2022-05-31 at 10 28 25 AM

I had to add some custom CSS to handle how the borders needed to look and interact with the hover states. I tried to get around it using other method (using the `peer` tailwind class, negative margins, etc) but couldn't come up with a way that wouldn't cause undesired artifacts to occur when hovering, like the following:

Screen Shot 2022-05-27 at 2 18 03 PM

Screen Shot 2022-05-27 at 2 10 06 PM

Screen Shot 2022-05-27 at 3 21 30 PM

I'm not sure if I handled this additional CSS in the preferred way. Let me know if I should make changes here or if there is a better way we prefer to handle this sort of thing. 

---
 
Closes #45

### DIFF
--- a/libs/design-system/src/lib/Components/LinkCard3Column/LinkCard3Column.stories.tsx
+++ b/libs/design-system/src/lib/Components/LinkCard3Column/LinkCard3Column.stories.tsx
@@ -1,0 +1,203 @@
+import { Meta, Story } from '@storybook/react';
+import { LinkCard3Column, LinkCard3ColumnProps } from '.';
+
+const Icon1 = () => (
+  <svg
+    width="38"
+    height="42"
+    viewBox="0 0 38 42"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M15.543 22.4783V18.8572L18.9988 16.9147L22.4545 18.8572V22.4783L18.9988 24.4208L15.543 22.4783Z"
+      stroke="#69717E"
+      stroke-width="0.5"
+    />
+    <path d="M16 22V19.5L18.5 21V23.5L16 22Z" fill="#69717E" stroke="#69717E" />
+    <path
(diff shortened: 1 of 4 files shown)


libs/design-system/src/lib/Components/LinkCard3Column/LinkCard3Column.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+export const Default = Template.bind({});
+Default.args = {
+  buttonText: 'Get Started',
+  buttonUrl: '#changeme',
+  description:
+    'Building on Flow is easy. Start building now with lorem ipsum et sigitus loranum prospitarius.',
+  title: 'Start Your Project',
+  tags: ['Tag', 'Lorem', 'Ipsum'],
+  items: [
+    {
```
10thfloor, line 120: Can we add an interface for these?

10thfloor (changes requested):
This looks great. One ask is to create an interface for the items in `interfaces.ts` that encodes the structure of the `item`(s) so we can use them in the real application.

alse:
Good idea adding the custom css for the borders @joekrill. Unfortunately the tailwind peer class applies to all following siblings instead of just the next one.

I renamed it to divided-item-hover, so we can use it with the divide-y helpers in other places: #82

alse (approved):
👍

libs/design-system/src/lib/Components/LinkCard3Column/LinkCard3Column.stories.tsx
10thfloor: I think anything that will be passed as a prop should be an interface, as in the props interface to the component" ... although I'm uncertain when you should use `type` vs `interface` ... Either way, as long as we can statically check our props I'm happy.

PR #82, opened 2022-06-01
Add divided-item-hover helper

alse:
.divided-item-hover hides borders on hover when using the tailwind divide helpers

```
<div className="divide-y divide-border-primary-gray-50">
   <div className="divided-item-hover">item</div>
   <div className="divided-item-hover">item</div>
   <div className="divided-item-hover">item</div>
</div>
```


aureentuluvacal (approved):
Changes look good to me. Nice addition!

joekrill (approved):
That's much better - thanks for cleaning things up!

PR #87, opened 2022-06-01
Updated Event Card

joekrill:
- Adds a generic `MobileCarousel` component that should be reusable in these cases where we want a vertical stack to become a horizontal carousel on mobile screens.
- Adds `EventCardList` component for rendering multiple events (uses the `MobileCarousel` component mentioned above)
- In `EventCard`: 
  - Makes the title clickable
  - Renames `buttonUrl` prop to `href`
  - Renames `buttonText` prop to `ctaText` (to match existing prop names)
  - Renames `eventType` prop to `location` (I think that's what this is meant to indicate?)

This is what the mobile version of the `EventCardList` looks like:

Screen Shot 2022-06-01 at 3 41 55 PM


alse (approved):
Very nice @joekrill , accessible too!

Issue #23, opened 2022-05-11
[Pages] Community


srinjoyc:
# Context

Create the Community page using the components library as per the design [here](https://www.figma.com/file/704SXr4HrUbIIkXYena62Q/Flow-Dev-Portal-(External)?node-id=1012%3A35901).

- [x] Header
- [x] Events
- [x] Contributors
- [x] FLIPs
- [x] Featured Initiatives
- [ ] Recent Tools
- [ ] Get Involved
- [x] Featured Article
- [ ] From the Forum
- [x] Subscribe

PR #111, opened 2022-06-06
Add contributors and project cards to Community page

alse:
#23

image
image


libs/design-system/src/lib/Pages/shared/GrayPage.tsx
```diff
@@ -0,0 +1,15 @@
+import clsx from 'clsx';
+
+export default function GrayPage({
```
aureentuluvacal, line 3: What do you think about renaming this to `PageBackground`?

aureentuluvacal (approved):
Looks fantastic. Thanks for adding the container class to the other components!

libs/design-system/src/lib/Pages/shared/GrayPage.tsx
alse: Sure, just renamed it!

PR #112, opened 2022-06-06
Learn page

joekrill:
Adds the "Learn" page

- There was already a "filter" icon but it was not the one used in the designs, so I added the necessary one and called it "filter2.svg" - not ideal but I didn't want to assume the other icon wasn't being used. We may want to rename this.
- The pagination is a little tricky because the mobile version only shows 4 items/page while the desktop version shows 12 items/page. To handle this dynamically we'd need to [import the current tailwind theme and "resolve" the current values](https://tailwindcss.com/docs/configuration#referencing-in-java-script) - but the tailwind config is stored outside of the design system and I wasn't sure if we could potentially be passing in different tailwind configs, and how this would resolve. It would still be tricky because we'd need to test the underlying media queries and watch for changes - and there's some problems with Safari's implementation. At the end of the day it just seemed simpler to render the thing content twice, with different page sizes, only showing the appropriate one. Happy to reconsider this implementation if we think it's going to be problematic.
- I modified the `Pagination` component so that it uses `<button/>` elements instead of divs (much better for a11y) and so that it doesn't allow moving to invalid pages.

localhost_4400_iframe html_id=pages-learnpage--default args= viewMode=story


libs/design-system/src/lib/Components/ToggleButton/ToggleButton.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+} as Meta;
+
+const Template: Story<ToggleButtonProps> = (args) => {
+  return (
+    <div
+      style={{
+        gap: '1rem',
+        padding: '1rem',
+        display: 'inline-flex',
+        flexDirection: 'column',
```
10thfloor, line 25: Do these have to be inline?

10thfloor (changes requested):

10thfloor (approved):

joekrill:
I've updated this to use the pagination as discussed yesterday:

- All 3 lists of tutorials are paginated
- All show 4 items at a time, regardless of screen size
- Removed "View tutorials" link from the NFTs section

Also:
- Extracted this logic to a component called `PaginatedTutorialCardList`
- Sort tags for consistency when displayed in "All content" section

libs/design-system/src/lib/Pages/LearnPage/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        : allTutorials,
+    [allTutorials, filters]
+  );
+
+  const tags = Array.from(
+    new Set(allTutorials.reduce((acc, { tags }) => [...acc, ...tags], []))
+  ).sort();
+
+  return (
+    <div className="container w-full bg-primary-gray-50 p-6 dark:bg-black">
```
alse, line 55: The container class will have to be moved to the individual page section components so that the gray background can be full width. I also added a <PageBackground> component that can replace this div

libs/design-system/src/lib/Components/TutorialCard/PaginatedTutorialCardList.tsx
```diff
@@ -0,0 +1,55 @@
...
+          .map((tutorialProps, index) => (
+            <TutorialCard key={index} className="w-full" {...tutorialProps} />
+          ))}
+      </div>
+      <Pagination
+        className="mt-4"
+        itemCount={tutorials.length}
+        page={page}
+        pageSize={pageSize}
+        setPage={setPage}
```
alse, line 49: Maybe this was planned for a future task, but it would be nice to scroll the user to the top of the section.

Currently, if the new page has less items, the user gets scrolled to the section below which is a bit confusing

libs/design-system/src/lib/Pages/LearnPage/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+          </ButtonLink>
+        </PageSection>
+
+        <PageSection className="flex-col items-stretch">
+          <h2 className="text-h2 f mb-6">NFTs</h2>
+          <PaginatedTutorialCardList tutorials={nftTutorials} />
+        </PageSection>
+
+        <PageSection>
+          <h2 className="text-h2 mb-6">Featured videos</h2>
```
alse, line 98: This needs a Go to YouTube button. I just added "external" and "right" icons to the button components that can be used here (https://github.com/onflow/next-docs-v1/pull/111/files#diff-74f71e61cdec7349c426ca754f9f12c3b91494f59e68f65fab35e2050a443964R16)

alse (changes requested):

libs/design-system/src/lib/Components/TutorialCard/PaginatedTutorialCardList.tsx
joekrill: That's a great point. I'll add this. It's a little trickier because I think we probably _don't_ want this behavior when changing the filters (which also resets the pagination to the first page). It also can feel a little weird in some cases -- in particular, on mobile when the list size is reduced because of navigating to the last page. We might be able to mitigate this by animating the cards in some way when the page changes. But I'll leave that for a later enhancement.

libs/design-system/src/lib/Pages/LearnPage/index.tsx
joekrill: Great catch. It looks like there's no "Go to Youtube" button on mobile, however. This mis probably an oversight since the other "header buttons" show up in mobile, so I'll make sure it's added there as well.

libs/design-system/src/lib/Components/TutorialCard/PaginatedTutorialCardList.tsx
alse: The scrolling looks great!

alse (approved):
👍

PR #120, opened 2022-06-07
Add prettier CI check to the remix site

schpet:


schpet:
@codingone21 @10thfloor is it an OK time to get this in? it

1. adds a CI check to make sure files are formatted
2. formats all of the files in the remix dir (i left out the other dirs to avoid creating a lot of merge conflicts for open PRs)

adding this check will make it easier to avoid git conflicts for me, since i frequently run prettier on entire files which may conflict if other folks do not.

PR #122, opened 2022-06-07
Refactor fetchFlips from json export/import to returning Props

codingone21:
- Changed implementation from fetching exported `.json` to fetching actual component format `FlipCellProps[]`. 
- Added a view page in routing, /flipsView for testing. 

Loaded View:
Screen Shot 2022-06-09 at 10 46 12 AM


schpet:
@codingone21 fyi i should be able to check this out once the build is working again! any instructions on how to run this code? e.g. do i go to localhost:3000/flipsView or something?

schpet:
@codingone21 fyi i should be able to check this out once the build is working again! any instructions on how to run this code? e.g. do i go to localhost:3000/flipsView or something?

codingone21:
> @codingone21 fyi i should be able to check this out once the build is working again! any instructions on how to run this code? e.g. do i go to localhost:3000/flipsView or something?

Thanks and yep! That should be the correct route. I'll also verify once build is working.

10thfloor (changes requested):
Nice. You can move this file into a `utils` folder as we discussed.

codingone21:
Successfully builds, but errors on loading due to how remix data loading works. LoaderFunction is meant to be use only under /routes. Working on a fix

10thfloor:
> Successfully builds, but errors on loading due to how remix data loading works. LoaderFunction is meant to be use only under /routes. Working on a fix

Ah so we can get reid of the LoaderFunction and just have this code in a generic function, that we can import anywhere we need (including but not necessarily a LoaderFunction)

codingone21:
> > Successfully builds, but errors on loading due to how remix data loading works. LoaderFunction is meant to be use only under /routes. Working on a fix
> 
> Ah so we can get rid of the LoaderFunction and just have this code in a generic function, that we can import anywhere we need (including but not necessarily a LoaderFunction)

Yep! exactly :) The recent commits changed the loader function to 'fetchFlips', which is just a generic call. We can call this from a loader function for components rendering

10thfloor (approved):

PR #124, opened 2022-06-08
Resolve build issues related to remix and stateful components

schpet:
**don't merge: not in a working state**

- import react!
- Switchup rollup config

### DIFF
--- a/libs/design-system/src/lib/Components/InternalCodeblock/index.tsx
+++ b/libs/design-system/src/lib/Components/InternalCodeblock/index.tsx
@@ -1,4 +1,3 @@
-import { Dialog } from '@reach/dialog';
 import '@reach/dialog/styles.css';
 import clsx from 'clsx';
 import React, { useState } from 'react';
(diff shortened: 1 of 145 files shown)


schpet:
thought this was working but unfortunately not the case

libs/design-system/src/lib/Components/InternalCodeblock/index.tsx
```diff
@@ -104,29 +103,6 @@ export function InternalCodeblock({
           )}
         />
       </div>
-      <Dialog
```
10thfloor, line 107: Was this removed intentionally?

10thfloor (approved):

libs/design-system/src/lib/Components/InternalCodeblock/index.tsx
schpet: yeah consider this whole thing an experiment that didn't work out. removing this changed the way it failed slightly: without dialog the page will load and errors are logged in the browser console, with the dialog the page doesn't load at all.

10thfloor:
Closing this. Upcoming work to remove NX and consolidate design-system into Remix app will fix this.

Issue #130, opened 2022-06-08
Permissionless banner


neilZon:
**Description**
We want to add a banner to the docs site with the following message, it should look pretty similar to this with the following sentence: "Permissionless deployment is coming to Flow! Read more [here](https://permissionless.onflow.org/).
![image](https://user-images.githubusercontent.com/46465568/172708602-bd568ac5-0f67-483a-8335-c61878b40942.png)

thread: https://axiomzen.slack.com/archives/C02SP5GS4GJ/p1654708067239369?thread_ts=1654707860.697179&cid=C02SP5GS4GJ

PR #131, opened 2022-06-08
Announcement Banner

neil-ptr:
**description**
Adds announcement banner

closes: #130

Screen Shot 2022-06-08 at 3 22 50 PM

### DIFF
--- a/app/ui/design-system/src/lib/Components/AnnouncementBanner/AnnouncementBanner.stories.tsx
+++ b/app/ui/design-system/src/lib/Components/AnnouncementBanner/AnnouncementBanner.stories.tsx
@@ -1,10 +1,10 @@
-import { Meta, Story } from "@storybook/react";
-import { AnnouncementBanner } from ".";
+import { Meta, Story } from "@storybook/react"
+import { AnnouncementBanner } from "."
 
 export default {
   component: AnnouncementBanner,
   title: "Components/AnnouncementBanner",
-} as Meta;
+} as Meta
 
 const Template: Story = (args) => {
   return (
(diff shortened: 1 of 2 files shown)


libs/design-system/src/lib/Components/AnnouncementBanner/AnnouncementBanner.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+  component: AnnouncementBanner,
+  title: 'Components/AnnouncementBanner',
+} as Meta;
+
+const Template: Story = (args) => {
+  return (
+    <AnnouncementBanner {...args}>
+      <div className="text-center text-xs font-semibold md:text-sm">
```
alse, line 12:
```suggestion
      <div className="text-center text-xs font-semibold md:text-sm dark:text-black">
```

alse, line 12: This needs a `dark:text-black` class. Maybe the styles should be in the AnnouncementBanner?

libs/design-system/src/lib/Components/AnnouncementBanner/AnnouncementBanner.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+        Permissionless deployment is coming to Flow! Read more{' '}
+        <span style={{ textDecoration: 'underline' }}>
+          <a href="https://permissionless.onflow.org/">here</a>
```
alse, line 15: Should this have a target="_blank" to open a new window? Assuming this component will be added to the site later?

alse (changes requested):

libs/design-system/src/lib/Components/AnnouncementBanner/AnnouncementBanner.stories.tsx
neilZon: Yeah that would a good idea, nice catch!

libs/design-system/src/lib/Components/AnnouncementBanner/AnnouncementBanner.stories.tsx
neilZon: Also a good catch!

10thfloor (approved):

alse (approved):
👍

PR #132, opened 2022-06-08
add component exports

neilZon:
**description**
Add component exports to `libs/design-system/index.ts`


aureentuluvacal:
Wait @neilZon is this supposed to export all the components?

neilZon:
@aureentuluvacal I only went back and did the ones I did just in case others went back and did theirs

10thfloor (approved):

PR #140, opened 2022-06-09
concept page component

neilZon:
**description**
Adds the concepts page

closes: #26


app/ui/design-system/src/lib/Pages/ConceptsPage/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+            />
+          ))}
+        </div>
+      </div>
+
+      <div className="pb-40 md:border-b md:border-primary-gray-100 md:pt-20 md:pb-28">
+        <div className="container">
+          <div className="flex justify-between md:pb-11">
+            <div>
+              <h4 className="text-h4 md:text-h2 pb-4 text-2xl md:pb-2">
```
alse, line 76: The tools section is the same as the one on the tools page, and very similar to the ToolsAndConcepts. What do you think about reusing the ToolsAndConcepts component in both places?

It can be modified to only show the tabs if both tools and concepts are passed as props, and take in custom props for the title/subtitle

app/ui/design-system/src/lib/Pages/ConceptsPage/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+export function ConceptsPage({
+  landingHeaderItems,
+  featureLinkBlockItems,
+  toolCardItems,
+  contentNavigationItems,
+  eventCardItems,
+}: ConceptsPageProps) {
+  return (
+    <div className="bg-primary-gray-50 dark:bg-black">
+      <div className="pb-10 md:border-b md:border-primary-gray-100 md:pb-28">
```
alse, line 45: Can the sections be changed to use the PageSections and PageSection components? Check out the Community page for an example

alse (changes requested):

app/ui/design-system/src/lib/Pages/ConceptsPage/index.tsx
alse: @neilZon looks like @aureentuluvacal already updated ToolsAndConcepts in #149 so it can be reused

neilZon:
> @neilZon looks like @aureentuluvacal already updated ToolsAndConcepts in #149 so it can be reused

@aureentuluvacal just to keep you in the loop I had to add some stuff to the ToolsAndConcepts to conform to the designs in Figma for the concept page!

10thfloor (approved):

PR #141, opened 2022-06-10
Add autoprefixer 10.4.7

alse:
postcss needs a newer autoprefixer to load docsearch.css


schpet (approved):
oh man so fast! thanks so much 🙏🏻

PR #142, opened 2022-06-10
Add Storybook dark mode decorator and adjust viewports

joekrill:
Adds the correct viewport values back to the Storybook configuration based on the tailwind screen size configurations.

Also attempts to fix "dark mode". This was previously handled by the `storybook-addon-themes` addon but that seems to not be working with the latest version of Storybook? Instead I've added a decorator that turns on dark mode when the "background" selection is set to dark. This is essentially the same behavior we had before with `storybook-addon-themes`, except:

1. Setting the parameter in a story to force dark mode no longer works in the way it previously did. But maybe this wasn't the best way to create stories in the first place? We weren't consistent with it across stories, anyway. 
2. I noted a bug where a hard-reload causes storybook to not properly report the current background to the underlying decorator. I tried various ways to get around this and was unsuccessful. The fix is just to toggle the background value again.


schpet (approved):
thanks so much for taking care of this!

10thfloor (approved):

alse:
Nice @joekrill ! I pushed some updates in #146 to set the root app classes in the preview app and enable dark mode stories. The dark mode stories are optional, but I like to add them as a reminder to check the dark versions

PR #155, opened 2022-06-13
Add network status fetching using API

codingone21:
app/cms/utils/hooks-network-status.tsx -> fetches network status by polling
app/routes/fetch-network.tsx -> temporary view for viewing network statuses

The returning data (StatuspageApiResponse[]) should be fed to the Network Card components
Screen Shot 2022-06-12 at 7 29 48 PM
.


app/cms/utils/hooks-network-status.tsx
```diff
@@ -0,0 +1,29 @@
...
+      refreshInterval: 100000,
+      onSuccess: (result) => {
+        console.log("setting from result")
+        setNetworkStatus(result)
+      },
+    }
+  )
+  console.log("network status", networkStatus)
+  return networkStatus as StatuspageApiResponse[]
+}
```
schpet, line 29: @codingone21 i would recommend avoiding the `swr` dependency in a remix app: instead recommend the loaders and `useFetcher` if you need to do things like polling. this is desirable for a few reason:

- you won't have CORS issues since the data will come from remix (i.e. the same origin), and remix loaders will make the request from it's backend
- data will be server rendered, not fetched after page load, which is in the spirit of remix and is faster


data fetching things like swr, react-query, graphql etc aren't needed in remix apps

schpet: i don't know if this is the best way to implement polling in remix, but something like this is what i'm trying to suggest:

https://benborgers.com/posts/remix-poll

app/routes/poll-network.tsx
```diff
@@ -0,0 +1,50 @@
+import { useFetcher, useLoaderData } from "@remix-run/react"
+import { useEffect, useState } from "react"
+import { fetchNetworkStatus } from "~/cms/utils/fetch-network-status"
+import { StatuspageApiResponse } from "~/ui/design-system/src/lib/interfaces"
+
+export async function loader() {
+  return await fetchNetworkStatus()
+}
```
schpet, line 8: i think it would probably be better to return an object here, e.g.

```
const networkStatus = await fetchNetworkStatus()
return {
  networkStatus
}
```

schpet (approved):
dropped some minor recommendations to consider! i'll also note that there's no built in error handling here, e.g. in case the fetch to STATUSPAGE_API_URL fails, but i think that's fine to leave out for now

PR #161, opened 2022-06-13
Add sidebar to repo pages

### LINKED ISSUE #61: [due June 20] Generate Menus from markdown content
For docs pages we'll need to generate the required (left and right hand side) menus dynamically as given by the designs.

- [ ] Menus are generated from markdown headings
- [ ] Handles the case where markdown page has no headings
- [ ] Menus are passed to correct layout and page components as props
- [ ] Menus are rendered correctly in the UI

schpet:
notes:

- introduces idea of a "repo schema" where things like sidebar data can be defined. right now it's a typescript file that's compiled into a [json-schema document](https://json-schema.org/), this will support validation, has nice typescript support, and type-ahead in IDEs like vs code to simplify authoring the json data. we could extend the repo schema to be useful for other data and configuration as well i think. eventually we could expose this online too, e.g. `https://whatever.onflow.com/repo-schema.json` so it can be more easily referenced.
- splits out repo and paths into separate routes: `app/routes/$repo.tsx` and `app/routes/$repo/$.tsx` – note the use of remix's `<Outlet>` here – this idea is a core part of remix and is a must to use it effectively
- existing sidebar data was found in the flow repo's gatsby config: https://cs.github.com/onflow/flow/blob/55844a8d449e73abdff81bb026f141bcda416dd3/docs/gatsby-config.js?q=%22Cadence+Design+Patterns%22#L340-L349

todo

- [ ] copy over menu data for repos other than cadence in a future PR (perhaps i can delegate this?)

Fixes #61 

Screen Shot 2022-06-14 at 1 21 14 PM
Screen Shot 2022-06-14 at 1 21 11 PM


app/constants/repos/repo-schema.json
```diff
@@ -0,0 +1,37 @@
...
+                    },
+                    "type": "array"
+                }
+            },
+            "type": "object"
+        }
+    },
+    "type": "object"
+}
+
```
schpet, line 37: later, we could make this available publicly on a remix route so that it could be referenced in a json document in the "$schema" field more easily

app/constants/repos/repo-schema.ts
```diff
@@ -0,0 +1,8 @@
+// run `yarn jsonschema` after making changes to the file
+// which updates repo-schema.json
+
+import { InternalSidebarConfig } from "~/ui/design-system/src/lib/Components/InternalSidebar"
+
+export type RepoSchema = {
+  sidebar: InternalSidebarConfig
+}
```
schpet, line 8: initially i went with a schema that was similar to docusaurus', but realized it would make more sense to use our existing types in the InternalSidebar component

app/routes/$repo.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const isKnownRepo = repoList.map((r) => r.repo).includes(params.repo)
+  if (!isKnownRepo) {
+    throw json({ status: "unknownRepo" }, { status: 404 })
+  }
+
+  // currently we are using only the presets, evnetually here we could
+  // use the github api to see if a configuration file (e.g. onflowdocs.json)
+  // exists, validate it against repo-schema, and use that instead. for now,
+  // we'll use the definitions in this repo
+  const sidebar = repoPresets[params.repo] ?? null
```
schpet, line 28: please note this is where we could see if a repo has their own sidebar config

app/routes/$repo.tsx
```diff
@@ -0,0 +1,74 @@
...
+export default function Repo() {
+  const data = useLoaderData<LoaderData>()
+  return (
+    <div className="mx-4">
+      <div className="flex">
+        <div>
+          {data.repoSchema ? (
+            <InternalSidebar config={data.repoSchema.sidebar} />
+          ) : (
+            <div>⚠️ Missing repo config</div>
```
schpet, line 45: right now, we only have cadence defined but once we have all the repos in maybe we can omit this branch

10thfloor (approved):

PR #169, opened 2022-06-14
Add homepage to remix app

alse:
#153

- Adds homepage to remix app
- Misc style updates


10thfloor:
Nice! 

A few things I noticed:


1) Are these placeholders? The quality is bad. :(
Screen Shot 2022-06-14 at 4 06 36 PM

2) The icons are clipping here.
We can apply `width` `height` and `viewBox` to the imported svgs if necessary.
Screen Shot 2022-06-14 at 4 06 43 PM

3) Dates are not formatted here: use `date-fns`
Screen Shot 2022-06-14 at 4 07 16 PM

4) There is a slight shadow on hover on the top "Card" but not the bottom one. They should both have a hover state?
Screen Shot 2022-06-14 at 4 03 07 PM

 5) Clicking on these tabs causes a 500 error. Not sure but i think this will require a deeper discussion of how we want these tabs to work with Remix routes / links
Screen Shot 2022-06-14 at 4 02 51 PM
 
That;s it for now, otherwise amazing to see this live!

10thfloor (changes requested):
See my comments in the main PR thread.

alse:
Thanks for the review @10thfloor , let me know if this needs any other changes!

1. Yes, those are placeholders that we'll have to update before launch
2. Good catch, it looks like some of the svgs need to be updated after the svgr changes. I fixed these and will keep an eye on any others
3. I added a helper function to format the dates, and we'll have to use the same method for all other dates in a future PR
4. I removed the shadow on the bottom one because only the 2 internal links are clickable, not the entire card. Should we add the hover state anyway, or remove it from both?
5. Yes, we'll have to figure out how we want to load the tabs. We could use links/routes, load them client side, or preload everything in the page loader

10thfloor (approved):

PR #172, opened 2022-06-14
Main Navigation

joekrill:
Implements the main navigation components. This turned out to be quite complex because the number variations is significant.  I'm certain there will need to be some adjustments to this. I had to take some liberties around the mobile design, dark mode, scrolling, and some of the more complex layout scenarios.

Desktop:

Kapture 2022-06-14 at 18 17 51

Mobile:

In addition to the main navigation, this includes a few additional minor changes:

- Removes the height/width from SVGs to prevent the viewBox from being stripped, which wasn't allowing us to scale SVGs.
- Adds `indicatorSize`, `indicatorColor`, and `indicatorSelectedColor` props to `Carousel`
- Prevents the tutorial card list from scrolling to itself on initial render


app/ui/design-system/src/lib/Components/LinkCard2Column/LinkCard2Column.stories.tsx
```diff
@@ -14,6 +14,7 @@ export const svgToDataUri = (
 export default {
   component: LinkCard2Column,
   title: "Components/LinkCard2Column",
+  excludeStories: ["svgToDataUri"],
```
10thfloor, line 17: Interesting. Is this something we'll need to do on the main site as well?

PR #173, opened 2022-06-14
Neil/tool card icon fix

neilZon:
# Description
Fixes issue where icons are clipping in the tool and SDK cards

## Bef
Screen Shot 2022-06-14 at 4 25 16 PM
ore

## After
Screen Shot 2022-06-14 at 4 27 15 PM


aureentuluvacal (dismissed):

app/ui/design-system/src/lib/Components/SDKCard/index.tsx
```diff
@@ -53,7 +53,12 @@ export function SDKCard({
           </div>
 
           <div className="flex shrink-0 gap-1">
-            <StarIcon className="h-fit text-amber-400" width={14} height={14} />
+            <StarIcon
+              className="text-amber-400"
+              viewBox="0 0 24 24"
```
joekrill, line 58: It feels strange to have to explicitly specify the `viewBox` here. If you remove the height/width from the underlying SVG (may need to run `yarn run dev:svg` after), svgr will keep the `viewBox` attribute and the SVG can then  be scaled by specifying height and/or width without needing to also specify `viewBox`. 

I wonder if we should consider configuring svgr to not do this in the first place? It's technically _svgo_ that is removing it -- the `removeViewBox` plugin is enabled by default, and svgr uses svgo under-the-hood.  @10thfloor what do you think? I think all of this changed recently with the repo reconfiguration because I don't think this was happening before.

neilZon:
removed `width` and `height` from underlying `.svg` files that are converted to react components by svgr as discussed in the doc site huddle svg meeting

neilZon:
Some updated screenshots
Screen Shot 2022-06-17 at 1 03 21 PM
Screen Shot 2022-06-17 at 1 03 32 PM
Screen Shot 2022-06-17 at 1 03 40 PM

10thfloor (approved):

PR #182, opened 2022-06-15
Integrate Main Navigation to Remix site.

10thfloor:


app/utils.ts
```diff
@@ -45,3 +46,10 @@ export function useMatchesData(
 export function validateEmail(email: unknown): email is string {
   return typeof email === "string" && email.length > 3 && email.includes("@")
 }
+
+export const svgToDataUri = (
+  element: Parameters<typeof renderToStaticMarkup>[0]
+) => {
+  const svgString = encodeURIComponent(renderToStaticMarkup(element))
+  return `data:image/svg+xml,${svgString}`
+}
```
schpet, line 55: is this used?

10thfloor, line 55: I'm going to leave this in, but I'm wondering why we need this? Should have raised this in the huddle.

app/ui/design-system/src/lib/Components/NavigationBar/ScrollLock.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    return () => {
+      documentElement.style.overflow = originalOverflow
+      documentElement.style.paddingRight = originalPaddingRight
+    }
+  }, [documentElement])
+}
+
+export const ScrollLock = () => {
+  useScrollLock()
```
schpet, line 41: given we never pass an argument to this fn, and on line 11 it bails without a documentElement, does this do anything?

schpet (approved):
this all lgtm but i am a bit confused how the ScrollLock component works

app/utils.ts
joekrill: I originally created this to be used in storybook stories, but I think it can now probably be removed and replaced with URLs.

joekrill:
> this all lgtm but i am a bit confused how the ScrollLock component works

It prevents the underlying document from being scrolled whenever a navigation popover is being shown. Otherwise if you use your scroll wheel the page will scroll underneath the nav in many scenarios. Here's a gif showing the behavior (ignore the other visual problems, I'm working on fixing those):

Kapture 2022-06-16 at 14 15 39

And the way it works is by adding an `overflow: hidden` to the document's `<html />` element. It also adds padding to account for any scroll bar that might have been shown (otherwise the page can reflow and jump around).

schpet:
@joekrill to clarify, i don't know how that code is run given we call `useScrollLock()`, and not, e.g. `useScrollLock(window.document)`

https://github.com/onflow/next-docs-v1/pull/182#discussion_r899343226

schpet:
@joekrill ahh nvm i see it assigns a default arg – i glossed over that thinking it was a ts type 🤦🏻

PR #183, opened 2022-06-16
Update fly.toml for production and staging deployments.

10thfloor:
Add action step for deploying to staging or production depending on which branch was updated. 

Both Fly environments have been established for the project. 

`main` -> Staging (via `fly.staging.toml`) @ flow-docs.fly.dev
`production` -> Production (via `fly.production.toml`) @ flow-docs-production.fly.dev


schpet (approved):
🚀

PR #188, opened 2022-06-16
Add good to start FLIP issues, Add tab change to the component

codingone21:
Changes:
* modified fetch-flip which now returns an object like so:
`return { openFlips, goodPlacesToStartFlips } as FlipsProps`
* added a react hook for tab changes `onTabChange={setSelectedTab}`between 0/1 for the Flips component

Screen Shot 2022-06-16 at 3 15 01 PM
Screen Shot 2022-06-16 at 3 15 07 PM


app/ui/design-system/src/lib/Components/Flips/index.tsx
```diff
@@ -30,17 +37,32 @@ export default function Flips({ flips }: FlipsProps) {
...
+          onTabChange={setSelectedTab}
         />
-      </div>
-
-      <FlipCellHeader />
-      <div className="flex flex-col gap-4">
-        {flips.map((flip, index) => (
-          <div key={index}>
-            <FlipCell {...flip} />
+        {selectedTab === 0 && (
```
alse, line 42: Setting the current tab's flips to a const could remove some duplication below

```
flips = selectedTab === 0 ? openFlips : goodPlacesToStartFlips
..
flips.map(..
```

alse (approved):
Left a small suggestion, but GTM!

PR #217, opened 2022-06-17
Feeding Forum Cell Components with Live Data

codingone21:
Feeding in Discourse API data into the Forum Cells, specifically the top 5 latest posts.

Screen Shot 2022-06-17 at 5 38 36 PM

Note for Future Work:
1. Currently, this component doesn't 'poll' every interval like originally planned. Instead, it fetches live data with every refresh. I ran into some errors for polling, and will be resolving them in the future.
2. We could improve the UI (like adding headers in the FLIPs cell and colouring)


10thfloor:
Nice!

app/cms/utils/fetch-discourse-api.tsx
```diff
@@ -50,28 +59,146 @@ export interface Topic {
...
+  posters: any
+}
+
+interface CategoryResponse {
+  category: Category
 }
 
-export interface Poster {
-  extras: string
+interface Category {
```
10thfloor, line 69: Oof this must have been painful. I wonder if there is a better way to do this?

10thfloor (approved):

app/cms/utils/fetch-discourse-api.tsx
codingone21: We can create an enum style mapping in our code as long as the category ID doesn't change!

PR #245, opened 2022-06-21
Cp/tab menu links

### LINKED ISSUE #229: Add link support to TabMenu
Continue work from #209 to allow links in the TabMenu

aureentuluvacal:
Closes #229


alse (changes requested):

app/routes/network/$networkName.tsx
```diff
@@ -20,7 +20,10 @@ export const loader: LoaderFunction = async () => {
 export default function Page() {
   const { networkStatuses } = useLoaderData<DynamicNetworkDetailPageProps>()
   const params = useParams()
-  const networkName = params.networkName as string
+
+  if (!params.networkName) throw new Error("Missing network name")
```
10thfloor, line 24: Do we need to throw here? If there is no network name we can assume it's a req for a different page, and this route doesn't get hit?

10thfloor (changes requested):

app/routes/network/$networkName.tsx
alse: I suggested this in another PR to avoid the type assertion, but I'm good with keeping it as it was

10thfloor (approved):

app/ui/design-system/src/lib/Components/TabMenu/index.tsx
```diff
@@ -39,16 +41,38 @@ const TabMenu = ({
...
-            id={name}
-            href={link}
-            className="relative cursor-pointer py-4 text-center text-black hover:text-primary-gray-400 dark:text-white hover:dark:text-primary-gray-100 md:py-6"
+            role="button"
             onClick={() => {
               setActiveIndex(index)
-              onTabChange(index, name)
             }}
+            className={tabClasses}
             isExternal={false}
```
10thfloor, line 76:
```suggestion
```

PR #276, opened 2022-06-21
Update discord invite

neilZon:
# Description
Update discord links to use discord URL constant


srinjoyc:
I think there's on in the footer and sidebar as well? Did we get those?

app/component-data/NavigationBar/joinFlowTabData.tsx
```diff
@@ -2,6 +2,7 @@ import EcosystemIcon from "~/ui/design-system/images/content/ecosystem"
 import AmbassadorsIcon from "~/ui/design-system/images/nav/ambassadors"
 import SocialsIcon from "~/ui/design-system/images/nav/socials"
 import { Section } from "~/ui/design-system/src/lib/Components/NavigationBar/types"
+import { DISCORD_URL } from "~/ui/design-system/src/lib/constants"
```
10thfloor, line 5: Constants that are part of the site should not be defined in the design system.
Move this to `/app/constants/index.ts`

10thfloor (changes requested):

srinjoyc:
Yep that's the one

neilZon:
@srinjoyc in that case it should be covered by the PR as well!

10thfloor (approved):

app/component-data/NavigationBar/joinFlowTabData.tsx
10thfloor: Ty!

Issue #299, opened 2022-06-23
Carousel Scroll Jumps


srinjoyc:
On the learn page, where we have carousels - when I click on the pagination, it forces a scroll jump - we should avoid doing this and keep the user in place.

srinjoyc:
The card sizes should be the same.

schpet:
FYI please see ongoing discussion https://github.com/onflow/next-docs-v1/pull/309#issuecomment-1165778655

PR #309, opened 2022-06-23
Only scroll on pagination of tutorial cards if necessary

joekrill:
This should reduce some of the scrolling behavior for the paginated tutorial cards. They should now only scroll when the first card in the list is not visible in the viewport. So on mobile we should see the same behavior as before, but on desktop we won't scroll down when the first card is in the middle of the viewport.   

@srinjoyc let me know what you think about this behavior.  A further option if this is still not ideal would be to check if _any_ part of the card (or even the whole collection) is visible, and not scroll in those cases. 

#299

### DIFF
--- a/app/ui/design-system/src/lib/Components/TutorialCard/PaginatedTutorialCardList.tsx
+++ b/app/ui/design-system/src/lib/Components/TutorialCard/PaginatedTutorialCardList.tsx
@@ -25,7 +25,7 @@ export const PaginatedTutorialCardList = ({
   tutorials,
   scrollOnPaginate = true,
 }: PaginatedTutorialCardListProps) => {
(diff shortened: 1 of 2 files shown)


10thfloor (approved):
Will merge once @srinjoyc approves.

srinjoyc:
@joekrill - its a lot better but if we could, there are two more improvements we can make:
- The size of the cards should be the same - so when scrolling, we can avoid another jump on the first row of learn page
- If we can add some smooth scroll animation on pagination that might make it less jittery as well


Let me know if this is dooable/makes sense.

schpet:
@srinjoyc FYI Joe and I discussed this, and decided it would be easier to get this merged and address your request in a follow up PR

PR #310, opened 2022-06-24
Fix right side menus that are generated from markdown headings

### LINKED ISSUE #192: Create navigation menus dynamically on internal pages
Work is already in progress here: https://github.com/onflow/next-docs-v1/pull/159

Create the navigation menus for all of the internal pages following the pattern established here: 
https://github.com/onflow/next-docs-v1/pull/161

schpet:
fixes #192, additionally

- Change dev:svg command to watch svg files for changes
- Adds jest

Screen Shot 2022-06-23 at 6 01 00 PM

Screen Shot 2022-06-23 at 6 01 03 PM


app/cms/utils/generate-toc.ts
```diff
@@ -1,15 +1,62 @@
...
+        hash: link.url,
+        title: linkText,
+      })
+    } else if (child.type === "list") {
+      items.push(...processItems(child.children))
+    }
+  }
+
+  return items
 }
```
schpet, line 62: fwiw this might not work for all markdown, but i feel like this is a good starting point and we can iterate on it

package.json
```diff
@@ -93,7 +97,7 @@
     "dev": "run-p dev:*",
     "dev:css": "yarn generate:css -- --watch",
     "dev:remix": "remix dev",
-    "dev:svg": "yarn svgr --silent --template svgr/template.js --filename-case kebab --out-dir ./app/ui/design-system/images ./app/ui/design-system/images",
+    "dev:svg": "chokidar --silent --initial \"./app/ui/design-system/images/**/*.svg\" -c \"yarn svgr --silent --template svgr/template.js --filename-case kebab --out-dir ./app/ui/design-system/images ./app/ui/design-system/images\"",
```
schpet, line 100: this makes it so that the svgs are "watched" and also let me run this app through a Procfile

joekrill (approved):
👍

app/ui/design-system/src/lib/Components/InternalToc/index.tsx
```diff
@@ -26,27 +26,22 @@ export function InternalToc({
...
-                    {title}
-                  </a>
-                </div>
-              )
-            })
-          : "No content"}
+        {headings.map(({ title, hash }, index) => (
+          <div className="flex" key={index}>
+            <a
+              href={hash}
```
alse, line 32: Should this be scrolling the page to the heading?

alse (approved):

app/ui/design-system/src/lib/Components/InternalToc/index.tsx
schpet: yeah that's what it does, `hash` here is a url anchor, e.g. `#publickey`

image

PR #311, opened 2022-06-24
Add display names for internal content/repos

joekrill:
This adds a "display name" for each piece of internal content so that we can use them in our breadcrumbs:

Screen Shot 2022-06-24 at 11 31 46 AM

By default the repo/content name is used,  replacing dashes with spaces and capitalizing each word. But each name can be overridden in the case this isn't ideal (for example "flow-cli" would by default be "Flow Cli", but we override this to "Flow CLI"). 

I reworked the repo constants a bit and tried to name things in a clear and distinguishable way, but I'm not really up-to-speed on a lot of terminology so let me know if there are better naming conventions here.


schpet (approved):
👍🏻

PR #315, opened 2022-06-24
Fix error loading cadence and other internal pages

schpet:
fixes issue introduced in 95c18027c9d198e88af8beae18589279d599ff8e


schpet:
fixed by #314

PR #317, opened 2022-06-24
For consideration: enable Typescript noUncheckedIndexedAccess option

joekrill:
Proposing we enabled `noUncheckedIndexedAccess` in our `tsconfig.json`. I've found this can catch a lot of bugs. I"m not entirely sure why this isn't the default these days (maybe for backwards compatibility?)

### DIFF
--- a/app/cms/utils/generate-toc.test.ts
+++ b/app/cms/utils/generate-toc.test.ts
@@ -76,7 +76,7 @@ test(`backticks`, () => {
 bar
   `
 
-  expect(markdownToToc(subject)![0].title).toBe(
+  expect(markdownToToc(subject)![0]!.title).toBe(
     `Avoid using AuthAccount as a function parameter`
   )
 })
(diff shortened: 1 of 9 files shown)


schpet:
wish they still called this "Pedantic index signatures" – IMO this would be better, but would recommend using invariants over the `!` assertions

joekrill:
> wish they still called this "Pedantic index signatures" – IMO this would be better, but would recommend using invariants over the `!` assertions

I think `!` assertions should really be limited to cases where you are _certain_ the thing is defined, but it's just not clear to Typescript. For example the `collection` variable in `app/ui/design-system/src/lib/Components/Search/Panel.tsx` _must_ be defined on line 51 because we are only at that point because `items` was defined, which came from `collection` in the first place. But that's also probably an indicator  that the code should be reworked in the first place. 🤷

10thfloor (approved):

Issue #282, opened 2022-06-21
Cadence Landing Page


10thfloor:
- Top level URL will be `/cadence`
- Needs special layout and menus
- Should be able to use component-data pattern to populate various UI elements, for now. 
- For insight into menu system ask @codingone21

PR #319, opened 2022-06-24
Add repo switcher to internal pages

### LINKED ISSUE #283: Implement repo switcher in remix site
image

schpet:
please note that i took the approach that worked with the existing `InternalSidebarMenu` implementation, it might make more sense to eventually change that component so that it doesn't have it's own concept of "ToolName" etc, but for now this was faster to implement.

additionally this loads index.md / index.mdx on paths like /cadence – i think it was working this way previously to some of my changes, but this change is relevant to #282 

Fixes #283 

Screen Shot 2022-06-24 at 1 18 25 PM
Screen Shot 2022-06-24 at 11 59 42 AM
Screen Shot 2022-06-24 at 1 19 10 PM

### DIFF
--- a/app/routes/$repo.tsx
+++ b/app/routes/$repo.tsx
@@ -7,8 +7,9 @@ import {
   useMatches,
 } from "@remix-run/react"
 import invariant from "tiny-invariant"
-import { ContentSpec, getContentSpec } from "~/constants/repos"
+import { ContentName, ContentSpec, getContentSpec } from "~/constants/repos"
 import { ErrorPage } from "~/ui/design-system/src/lib/Components/ErrorPage"
+import { ToolName } from "~/ui/design-system/src/lib/Components/Internal/tools"
 import { temporarilyRedirectToComingSoon } from "~/utils/features"
 import { InternalPage } from "../ui/design-system/src/lib/Pages/InternalPage"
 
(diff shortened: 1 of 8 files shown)


app/routes/$repo.tsx
```diff
@@ -49,12 +50,39 @@ export default function Repo() {
...
+  testing: "flow-js-testing",
+  vscode: "vscode-extension",
+}
+
+const toolLinks: Record<ToolName, string> = { ...toolContentMap }
+for (let [key, value] of Object.entries(toolLinks)) {
+  toolLinks[key as ToolName] = `/${value}`
+}
+
+const contentToolMap: Record<string, ToolName> = {
```
joekrill, line 77: Are these just reversed mappings of each other? Are they meant to diverge at any point? Could be prone to errors, or updating one and not the other.  Since `toolContentMap` seems to only be used for generating the `toolLinks`, would it be simpler to just generate that from `contentToolMap` so there's only one place to specify the mapping? Something like this maybe?

```ts
const toolLinks = Object.entries(contentToolMap).map(([content, tool]) => ({ [tool]: `/${content}` }))
```

joekrill (approved):

app/routes/$repo.tsx
schpet: agree that would be better, but IMO the core issue is that this mapping needs to exist at all so i don't think i'm going to bother with the small improvements. IMO al l the tools (and their images etc) should be passed in as props

PR #320, opened 2022-06-24
Increase NavigationBar z-index

joekrill:
Increases the navigation bar z-index. It needs to be higher than the breadcrumbs. It should be pretty high in the order (`z-40`), but probably not the max (`z-50`) which should probably be reserved for things like modals.


schpet (approved):
🤙🏻

PR #322, opened 2022-06-24
Add cadence landing page header

### LINKED ISSUE #304: [due July 5] Implement Cadence Landing page
As per the design, let's design the Cadence landing page - I believe we are going to be using the left hand layout but everything else is likely going to be custom? TBD.

joekrill:
Adds a header to the cadence landing page

Screen Shot 2022-06-24 at 3 26 17 PM

The props for this are specified by the same way the sidebar schema is. So this will make it simpler fi we want to allow the content authors to define/modify these headers. Seemed liked the best approach, but open to reworking if there are suggestions on how to better approach this.

---

Closes #304

### DIFF
--- a/app/routes/$repo.tsx
+++ b/app/routes/$repo.tsx
@@ -43,12 +43,14 @@ export default function Repo() {
   const { content } = useLoaderData<LoaderData>()
   const matches = useMatches()
   const [match] = matches.slice(-1)
+  const path = match.params["*"] || "index"
 
   return (
     <InternalPage
-      activePath={match.params["*"] || "index"}
+      activePath={path}
       contentDisplayName={content.displayName}
       contentPath={content.contentName}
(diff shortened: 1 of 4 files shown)


app/routes/$repo.tsx
```diff
@@ -42,12 +42,14 @@ export default function Repo() {
...
   const [match] = matches.slice(-1)
+  const path = match.params["*"] || "index"
 
   return (
     <InternalPage
-      activePath={match.params["*"] || "index"}
+      activePath={path}
       contentDisplayName={content.displayName}
       contentPath={content.contentName}
+      header={path === "index" ? content.landingHeader : undefined}
```
schpet, line 52: can you please merge in what's on `main` and see if this works when you go to `http://localhost:3000/cadence` ? might need to handle a null/blank/undefined path

schpet (approved):

PR #313, opened 2022-06-24
Add network status to navbar

aureentuluvacal:
Closes #302

PR #337, opened 2022-06-27
Improve behavior of getRequiredEnvVarFromObj

schpet:
instead of automatically providing an env var in dev for every api key, which doesn't always make sense, this can be done on an as-needed-basis


schpet:
@aureentuluvacal i was trying to figure out why my logs had nothing when i looked at #313 – this fixes that!

PR #340, opened 2022-06-27
Adds "Next" and "Previous" buttons  to internal pages

### LINKED ISSUE #332: Next and previous page buttons
image
Similar to the existing docs site, we should have next and previous buttons on the bottom of internal pages.

The logic should be either user-defined (ideally via the navigation JSON) or we could just do the next/prev file (if available) in the folder.

joekrill:
Adds the `LowerPageNav` component to internal pages.

Example (ignore the processing error):

Screen Shot 2022-06-27 at 3 04 17 PM

Closes #332


schpet (approved):
lovely! 👍🏻

PR #355, opened 2022-06-28
Show "Edit on Github" link for internal pages

joekrill:
This ultimately enables the "Edit on Github" link shown on internal pages:

Screen Shot 2022-06-28 at 5 06 09 PM

In order to do this, we can no longer render the <InternalPage /> until we actual fetch the data from github, because we won't know the link until we pull the content (it might be a directory, it might the root `index.mdx` file, or it might be a file itself and we just need to add the `.mdx` extension). So the `app/routes/$repo.tsx` file is no longer needed because everything needed to be hoisted up into the `app/routes/$repo/$.tsx` file.

Further, we need to know which branch we're working with, so that needed to be added to the `ContentSpec` (it defaults to `master`). 

I also pulled the `owner` out into the `ContentSpec` for consistency, and reworked the `getMdx()` method and it's underlying dependencies to take both an `owner` and `branch` value (with one exception: `downloadFileBySha` is branch agnostic). 

I think this ultimately simplifies things a bit, because now all of the repo/content information is specified in the `app/constants/repos/index.ts` file, and most of the methods no longer need to know whether something is "flow content", or how to piece together the path, etc.


app/cms/utils/mdx.tsx
```diff
@@ -209,13 +255,21 @@ async function compileMdxCached({
...
           ...compiledPage,
           fileOrDirPath,
+          editLink: [
+            "https://github.com",
+            owner,
+            repo,
+            "blob",
+            branch,
+            entry,
+          ].join("/"),
```
schpet, line 272: branch and entry always url safe, or is something like `encodeURIComponent` needed?

schpet (approved):
i'm sad to lose the nested `$repo` / `$repo/$.tsx` nested routing but can appreciate how the design doesn't really support it, making this change helpful

PR #357, opened 2022-06-28
Avoid referencing process in client code

schpet:
Screen Shot 2022-06-28 at 4 38 16 PM


schpet:
fyi @joekrill i _think_ this file is imported in client code, where process causes issues. i'm not sure if this repo owner needs to be configurable but i imagine this is a safe change

PR #362, opened 2022-06-29
Narrow `colorette` resolution

joekrill:
It looks like webpack-dev-server uses `colorette` as well, so changing the resolution globally broke Storybook. I think this narrows it correctly to only resolve that specific version for `@redocly/openapi-core`. @aureentuluvacal can you verify this looks right? It makes Storybook work correctly but I'm not totally sure how to test the its use with `redoc`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,6 @@
     }
   },
   "resolutions": {
-    "colorette": "2.0.19"
+    "redoc/**/colorette": "2.0.19"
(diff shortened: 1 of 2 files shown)


schpet:
@joekrill were you able to visit `/http-api` with this change? should behave like this:
https://flow-docs-staging.fly.dev/http-api

PR #363, opened 2022-06-29
Add mdx jsx support

### LINKED ISSUE #344: MDX component support
it appears that MDX components are not currently supported 

1. note the `<Callout>` mdx component is used in cadence anti patterns with the text "This anti-pattern has been addressed with [FLIP #703]…" https://github.com/onflow/cadence/blame/5a06fec82b4ab90d3fd770abb6b75ab56f2b51cd/docs/anti-patterns.mdx#L185
1. visit https://flow-docs-staging.fly.dev/cadence/anti-patterns
1. note that the component isn't shown, and the "This anti-pattern has been addressed" text is not present

aureentuluvacal:
Closes #344 

So it's literally one option 😂, but I'm uncertain of the security implications since this allows all JSX to pass through the sanitizer. All the normal HTML tags are still sanitized though since rehype sanitize is untouched.


joekrill:
Is the `yarn.lock` change intentional? I don't see any changes to the `package.json` and it doesn't look like it's just an upgrade.

aureentuluvacal:
> Is the `yarn.lock` change intentional? I don't see any changes to the `package.json` and it doesn't look like it's just an upgrade.

Lemme clean that up

Issue #361, opened 2022-06-29
MD Components UI Fixes


srinjoyc:
The following elements don't render correctly:

- [x] Bullets in list should be black instead of light grey (or if design is grey, we should try a much darker grey)
- [x] Codeblocks like `Node version v12.0.0 or higher.` with a single backtick are not rendering as that. Ideally we should have some special highlighting on these and not show the backticks (eg: Node version v12.0.0 or higher.)
- [x] External links are double underlined which doesn't look too good - we should ideally still have the arrow but not the extra line and should try to space out the next element after the arrow (see img)
- [x] Bottom prev and next should be within the same margins as the internal page
- [ ] Emoji's (github supported) does not render ie. 📢  or 🎉 
- [x] Videos are not showing up: https://flow-docs-staging.fly.dev/fcl-js/tutorials/flow-app-quickstart#video-walkthrough
![image](https://user-images.githubusercontent.com/8709330/176467431-8f67ce38-a239-4097-923b-0d57d4740edf.png)
![image](https://user-images.githubusercontent.com/8709330/176467696-b0548c54-5bdb-43cb-ac3b-d09dadb4d6da.png)
- [x] Fix heading styles when rendered from MDX.

srinjoyc:
The bullets are rendering a bit odd:
![image](https://user-images.githubusercontent.com/8709330/176719812-ab25442f-c94c-4da7-9aff-d152830ebee2.png)

PR #364, opened 2022-06-29
MD Components UI Fixes - Part 1

joekrill:
These are a few styling fixes that are part of #361:

- Changes color used for bullets in lists
- Styles inline codeblocks according to style guide
- Adjusts external link styling (removes "double-underline" and adjusts the icon)

### DIFF
--- a/app/ui/design-system/src/lib/Components/Link/ExternalLinkIcon.tsx
+++ b/app/ui/design-system/src/lib/Components/Link/ExternalLinkIcon.tsx
@@ -2,8 +2,8 @@ function ExternalLinkIcon() {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
-      width="19"
-      height="18"
+      width="1em"
+      height="1em"
       fill="none"
       viewBox="0 0 19 18"
     >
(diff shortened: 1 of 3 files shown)


alse (approved):

alse:
@joekrill the inline code looks good in paragraphs, but the styles don't get picked up in headings on /cadence/anti-patterns

image


mdx.css also had some outdated code styling. It wasn't showing due to a missing .mdx-content class. I removed it [here](https://github.com/onflow/next-docs-v1/pull/366/files#diff-744eababf13c4ced18cd5980bee47cf4357bf97261342030fdde167f8b5933f4L81) to avoid conflicts

joekrill:
> @joekrill the inline code looks good in paragraphs, but the styles don't get picked up in headings on /cadence/anti-patterns

It looks like this is happening because `h#` elements are explicitly given the `not-prose` prose in any markdown document:

Screen Shot 2022-06-30 at 9 26 07 AM
 
I'm not sure why this is the case. Any Idea?

joekrill:
@alse I think I see why this was done now (to facilitate the "floating" link on the left of headings). I'm going to deal with those in a separate PR - I added a separate bullet point to the issue.

Issue #360, opened 2022-06-29
Codeblocks UI


srinjoyc:
- [x] Codeblocks should default to the height of the code to a max of the viewport - roughly code sample pasted below should be visible w/o scrolling on most desktop screens.

- [x] We need to apply some default coloring to the codeblocks somehow. Cadence can be Swift highlighting for now, but most other code blocks that have ```js ``` should use JS, etc.

- [x] Remove this icon as it looks too close to the clipboard icon 
![image](https://user-images.githubusercontent.com/8709330/176468809-039cf513-2712-479a-a6a9-965bfd1e1d25.png)

```
import Head from 'next/head'
import "../flow/config";
import { useState, useEffect } from "react";
import * as fcl from "@onflow/fcl";

export default function Home() {

  const [user, setUser] = useState({loggedIn: null})

  useEffect(() => fcl.currentUser.subscribe(setUser), [])

  const AuthedState = () => {
    return (
      <div>
        <div>Address: {user?.addr ?? "No Address"}</div>
        <button onClick={fcl.unauthenticate}>Log Out</button>
      </div>
    )
  }

  const UnauthenticatedState = () => {
    return (
      <div>
        <button onClick={fcl.logIn}>Log In</button>
        <button onClick={fcl.signUp}>Sign Up</button>
      </div>
    )
  }

  return (
    <div>
      <Head>
        <title>FCL Quickstart with NextJS</title>
        <meta name="description" content="My first web3 app on Flow!" />
        <link rel="icon" href="/favicon.png" />
      </Head>
      <h1>Flow App</h1>
      {user.loggedIn
        ? <AuthedState />
        : <UnauthenticatedState />
      }
    </div>
  );
}
```

PR #366, opened 2022-06-30
Code block updates

alse:
#360

### DIFF
--- a/app/cms/compile.mdx.server.ts
+++ b/app/cms/compile.mdx.server.ts
@@ -5,12 +5,12 @@ import { bundleMDX } from "mdx-bundler"
 import type TPQueue from "p-queue"
 import path from "path"
 import calculateReadingTime from "reading-time"
+import rehypeSanitize, { defaultSchema } from "rehype-sanitize"
 import type * as U from "unified"
 import { visit } from "unist-util-visit"
+import type { GitHubFile } from "./github.server"
 import formatLinks from "./utils/format-links"
 import { markdownToToc } from "./utils/generate-toc"
-import type { GitHubFile } from "./github.server"
-import rehypeSanitize from "rehype-sanitize"
 
 if (process.platform === "win32") {
   process.env.ESBUILD_BINARY_PATH = path.resolve(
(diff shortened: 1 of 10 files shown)


app/cms/compile.mdx.server.ts
```diff
@@ -57,7 +57,16 @@ const remarkPlugins: U.PluggableList = [
...
   removePreContainerDivs,
   formatLinks,
-  rehypeSanitize,
+  [
+    rehypeSanitize,
+    {
+      ...defaultSchema,
+      attributes: {
+        ...defaultSchema.attributes,
+        code: [...(defaultSchema?.attributes?.code || []), ["className"]],
```
alse, line 66: @schpet I enabled className props on code blocks to enable code highlighting

joekrill (approved):

app/ui/design-system/src/lib/Components/InternalCodeblock/index.tsx
```diff
@@ -86,10 +86,10 @@ export function InternalCodeblock({
         <Code
           language={children.props.className}
           code={codeString}
-          innerClasses={clsx(
-            "w-full",
-            tall ? "max-h-[280px]" : "max-h-[130px]"
-          )}
+          innerClasses={clsx("w-full")}
+          innerStyle={{
+            maxHeight: constrained ? 280 : `calc(100vh - ${NAV_BARS_HEIGHT}px)`,
```
joekrill, line 91: I don't think this is _ideal_ because it assumes an `InternalCodeblock` is always going to be rendered in the context of a `NavigationBar` and `InternalSubnav`. Plus those heights could easily change, and they aren't really guaranteed as-is. But since we're already using that elsewhere it works for now.

alse: @joekrill the goal was to just maximize the code block height on taller windows, and I wanted to avoid querying the dom for the parent's height. The nav height is fixed using the same const, and there's a buffer to give it some extra space.

Good point about assuming this will always be rendered on internal pages, I changed the default state to a fixed height in #367

Thanks for reviewing!

PR #370, opened 2022-06-30
Put repo in relative link paths

### LINKED ISSUE #369: Relative links not replaced correctly
Found some links that did not replace correctly:
image

https://flow-docs-staging.fly.dev/fcl-js/tutorials/flow-app-quickstart

The link is: https://flow-docs-staging.fly.dev/reference/api where it should be https://flow-docs-staging.fly.dev/fcl-js/reference/api

neilZon:
# Description
This PR fixes the issue where the relative links are missing the root repo in their path
.../reference/api where it should be .../**fcl-js**/reference/api

closes #369

### DIFF
--- a/app/cms/utils/format-links.ts
+++ b/app/cms/utils/format-links.ts
@@ -16,8 +16,7 @@ function updateRelativeDepth(linkText: string, index = false) {
 
 module.exports = updateRelativeDepth
 
-// @ts-expect-error: TODO: Needs parameter types.
-const formatLinks = (...args) => {
+const formatLinks = (repoName: string) => {
   return async function rewriteRelativeLinks(tree: H.Root) {
     visit(
       tree,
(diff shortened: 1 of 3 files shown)


app/cms/utils/format-links.ts
```diff
@@ -36,6 +35,9 @@ const formatLinks = (...args) => {
 
           href = href.replace(/(?<=[^/])#/, "/#")
           href = href.replace(".mdx", "").replace(".md", "")
+          // remove parent or current directory paths from href, ie. remove: `../` or `./`
```
alse, line 38: @neilZon @srinjoyc I'm assuming there are no pages more than 1 level deep, or no repos that contain multiple docs folders linking to each other?

10thfloor: We should account for pages at an arbitrary depth. This will require either some redundancy or clever coding.

neilZon: This comment might be a bit misleading but it should say `removes all occurrences of the pattern '../' or './'`. But now that you mention it, this solution assumes that the lowest common ancestor is the docs folder, but it could be that the lowest common ancestor is lower in the directory tree when 1 file is giving a relative path to another so this will require a bit more thought

10thfloor (approved):

10thfloor (approved):

app/cms/utils/format-links.ts
10thfloor: Actually, we'll leave `docs` as the lowest common ancestor. This works!

Issue #368, opened 2022-06-30
Internal Page Sidebar & Mobile UI


srinjoyc:
The right sidebar should be collapsable - ideally on both mobile and web but definitely on mobile to allow the content to be displayed. There should be a sticky arrow that allows you to expand the bar to fill out the view on mobile to navigate.

srinjoyc:
Use https://docusaurus.io/docs/versioning as a reference for collapsible and mobile views for behavior.

Left and right sidebar will stack on top of content

PR #371, opened 2022-07-01
Update internal page layout

joekrill:
This updates and simplifies the `InternalPage` layout to the desired design for desktop. The heights of the sidebars are now calculated based on the actual height/position of the navbar as opposed to hard-coding expected heights. 

Also adds a few utilities:

- `canUseDOM` can be used to determine if the DOM is available (i.e. are we server-side rendering?)
- `useIsomorphicLayoutEffect` a `useLayoutEffect` replacement that can be safely used when server-side rendering.
- `useResizeObserver` provides the ability to watch an element for resize events in a performant way.

Will follow this up with mobile features as described in #368

### DIFF
--- a/app/routes/$repo/$.tsx
+++ b/app/routes/$repo/$.tsx
@@ -1,6 +1,5 @@
 import { json, LoaderFunction, redirect } from "@remix-run/node"
 import { Link, useCatch, useLoaderData, useLocation } from "@remix-run/react"
-import { useEffect, useState } from "react"
 import invariant from "tiny-invariant"
 import { getMdxPage, useMdxComponent } from "~/cms/utils/mdx"
 import {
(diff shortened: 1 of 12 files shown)


aureentuluvacal (approved):
This is awesome. I think instead of using the word "shared" for the resize observer, you could use "singleton". That's totally a nitpick though.

10thfloor (changes requested):
This looks great. I think we should move the new hooks out of the design system, and into the main app.

joekrill:
> This looks great. I think we should move the new hooks out of the design system, and into the main app.

I'm not totally sure how this would work because the design system components have direct dependencies on those hooks. So we'd then need to introduce some sort of dependency injection or something similar to make them available to the underlying components. They're also fairly generic -- they're not really specific to the app code.

schpet (approved):
lgtm 👍🏻

schpet:
@10thfloor re-requesting your review here given [Joe's comment](https://github.com/onflow/next-docs-v1/pull/371#issuecomment-1175022523)

10thfloor (approved):

Issue #347, opened 2022-06-28
MDX Error pages


neilZon:
# Description
The following pages receive a 500 error with the message "mdxError"

related 292

Affected pages:

- [ ]  /fcl-js/reference/api
- [ ]  http://localhost:3000/flow/content/concepts/transaction-lifecycle
- [ ]  http://localhost:3000/flow/content/concepts/variable-transaction-fees
- [ ]  http://localhost:3000/flow/content/dapp-development/dapp-infrastructure/deployment
- [ ]  http://localhost:3000/flow/content/dapp-development/dapp-infrastructure/flow-dapp-anatomy
- [ ]  http://localhost:3000/flow/content/dapp-development/dapp-infrastructure/in-dapp-payments
- [ ]  http://localhost:3000/flow/content/dapp-development/smart-contracts
- [ ]  http://localhost:3000/flow/content/flow-token
- [ ]  http://localhost:3000/flow/content/flow-token/wallets
- [ ]  http://localhost:3000/flow/content/intro/flow-token
- [ ]  http://localhost:3000/flow/content/node-operation/leased-account-setup
- [ ]  http://localhost:3000/flow/content/node-operation/node-roles
- [ ]  http://localhost:3000/flow/content/node-operation/node-setup
- [ ]  http://localhost:3000/flow/content/node-operation/guides/genesis-bootstrap
- [ ]  http://localhost:3000/flow/content/staking/staking-scripts-events
- [ ]  http://localhost:3000/flow/content/access-api
- [ ]  http://localhost:3000/flow/content

aureentuluvacal:
Hit a roadblock documented here https://tanookilabs.slack.com/archives/C03D8F2T68H/p1656696489983109?thread_ts=1656691162.732799&cid=C03D8F2T68H

Issue #292, opened 2022-06-22
Broken Internal Pages


neilZon:
# Description
Some of the internal pages have errors when you try to access them and this is a list of what pages have what errors

~~**Note 1** This only tests the `cadence/...` pages, this list is probably gonna get bigger~~
**Note 2** atm, we have a redirect to a "coming soon" if you try to access a page that isn't one of the landing pages so you won't be able to see the page errors if you click on the links below
<br/>

## Repos Tested So Far

-   cadence
-   fcl-js
-   flow-go-sdk
-   flow-js-testing
-   Flow/content
      - bounties
      - community-updates
      - concepts
      - core-contracts
      - faq
      - emulator
      - faq
      - flow-port
      - flow-token
      - fusd
      - intro
      - kitty-items
      - nft-marketplace
      - node-operation
      - staking
      - vscode-extension
 - flow-cadut
 - flow-nft
 - flow-ft
 - flow-emulator
 - nft-storefront
<br/>

### Callout Issues
Console: 
```
Error: Expected component `Callout` to be defined: you likely forgot to import, pass, or provide it.
```

- [x] https://developers.onflow.org/cadence/anti-patterns
- [x] https://developers.onflow.org/cadence/language/accounts
- [x] https://developers.onflow.org/cadence/language/built-in-functions
- [x] https://developers.onflow.org/cadence/language/composite-types
- [x] https://developers.onflow.org/cadence/language/resources
- [x] https://developers.onflow.org/cadence/language/interfaces
- [x] https://developers.onflow.org/cadence/language/values-and-types

### MDX Render Error #347 

- [x] /fcl-js/reference/api
- [x] http://localhost:3000/flow/content/concepts/transaction-lifecycle
- [x] http://localhost:3000/flow/content/concepts/variable-transaction-fees
- [ ] http://localhost:3000/flow/content/dapp-development/dapp-infrastructure/deployment
- [ ] http://localhost:3000/flow/content/dapp-development/dapp-infrastructure/flow-dapp-anatomy
- [ ] http://localhost:3000/flow/content/dapp-development/dapp-infrastructure/in-dapp-payments
- [x] http://localhost:3000/flow/content/dapp-development/smart-contracts
- [x] http://localhost:3000/flow/content/flow-token
- [x] http://localhost:3000/flow/content/flow-token/wallets
- [x] http://localhost:3000/flow/content/intro/flow-token
- [x] http://localhost:3000/flow/content/node-operation/leased-account-setup
[...]

PR #375, opened 2022-07-04
Fix internal pages from collected lists.

10thfloor:
Use the following issues to fix up internal pages. 
The checklists on these issues should be up to date, with which issues are fixed. 
If you fix a page, please check the box in the corresponding issue.
If you find a page already fixed, please chack the box as well.

Issues here: 
#292 
#388 
#347


10thfloor:
This work was completed.

PR #387, opened 2022-07-06
Allow trusted repos that skip rehypeSanitize

joekrill:
The `rehypeSanitize` plugin strips anything that it doesn't recognize, which includes any JSX components. These applies to any iframe videos and is causing the video not to show up for the `fcl-js/tutorials/flow-app-quickstart` page. The only way I've been able to get around this is to disable the `rehypeSanitize` plugin all-together. We obviously only want to do that for trusted sources so this allows setting an `isTrusted` property on on the `ContentSpec` which ultimately disables the sanitization. Right now this is set to trust all current repos, but should be adjusted as needed:

```
// app/constants/repos/index.ts
const trustedRepositories = [...repositoryNames];
```

### DIFF
--- a/app/cms/compile.mdx.server.ts
+++ b/app/cms/compile.mdx.server.ts
@@ -55,11 +55,17 @@ const remarkPlugins: U.PluggableList = [
   ],
 ]
 
-const rehypePlugins = (repoName: string): U.PluggableList => {
-  return [
+const rehypePlugins = (
+  repoName: string,
+  isTrusted: boolean = false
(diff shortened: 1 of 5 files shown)


10thfloor:
I'll merge this to unblock us for now, but since the onflow repos receive contributions from the public, we'll need to come back to this and figure out how to sanitize things in the correct order so we're not nuking jsx. I think @alse found a solution we can explore.

10thfloor (approved):

PR #392, opened 2022-07-06
Allow main layout to grow to full width

joekrill:
Fixes the main layout so it grows to the full width of the viewport.

Also tweaks the position of the main nav dropdowns - they were a little too low and allowed the top of the scrollbars for the main content to be visible.

### DIFF
--- a/app/ui/design-system/src/lib/Components/NavigationBar/DesktopMenuItem.tsx
+++ b/app/ui/design-system/src/lib/Components/NavigationBar/DesktopMenuItem.tsx
@@ -1,6 +1,7 @@
 import { Popover, Transition } from "@headlessui/react"
 import clsx from "clsx"
 import { Fragment } from "react"
+import { NAV_HEIGHT } from "."
 import { ReactComponent as ChevronDown } from "../../../../images/arrows/chevron-down"
 import { DesktopMenuTabbed } from "./DesktopMenuTabbed"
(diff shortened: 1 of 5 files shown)


app/ui/design-system/src/lib/Components/NavigationBar/DesktopMenuItem.tsx
```diff
@@ -61,15 +61,15 @@ export function DesktopMenuItem({ divider, ...props }: DesktopMenuItemProps) {
             leaveFrom="transform opacity-100 scale-100"
             leaveTo="transform opacity-0 scale-95"
           >
-            <Popover.Panel className="fixed top-[96px] left-0 right-0 bottom-0 z-40 origin-top-right">
+            <Popover.Panel className="fixed top-[90px] left-0 right-0 bottom-0 z-40 origin-top-right">
```
alse, line 64: Can these use the `NAV_HEIGHT` constant so they stay in sync? It may need to be updated to 90px

alse (approved):

app/ui/design-system/src/lib/Components/NavigationBar/DesktopMenuItem.tsx
joekrill: Ah didn't even realize that existed. Will update that.

PR #398, opened 2022-07-06
Support switching between versions of docs

schpet:
status

- [x] basic loading and switching between of versions with hardcoded versions
- [x] loading versions from the versions.json file ([example](https://github.com/onflow/mock-developer-doc/blob/81115de6e69db42b393704aa5cc11cec2f9ae6b0/versions.json)) instead of a hardcoded list
- [ ] use a polished UI for the version switcher
- [ ] support loading versioned images/files 
- [ ] determine what to do about the sidebars: we need versioned sidebars for this to make sense and be navigate-able
- [ ] if there is a versioned docs, i.e. `versioned_docs/version-1.0.0/` show that one by default instead of whatever is in `docs/`
- [ ] show the most recent version instead of "latest" when there are versions


10thfloor:
Closing as stale.

PR #404, opened 2022-07-07
Add an ORIGIN environment variable for links

schpet:
useful for sitemap, and places where we need to link to a fully qualified url and not a path

also exposed on window.ENV: 

Screen Shot 2022-07-07 at 12 03 50 PM

### DIFF
--- a/app/env.server.ts
+++ b/app/env.server.ts
@@ -16,13 +16,29 @@ function getEnv() {
 
 type ENV = ReturnType<typeof getEnv>
 
-// App puts these on
-declare global {
-  // eslint-disable-next-line
-  var ENV: ENV
-  interface Window {
-    ENV: ENV
+/**
+ * these values are exposed on window, don't put secrets here
+ */
+function getPublicEnv() {
+  return {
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -91,7 +91,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: superfly/flyctl-actions/setup-flyctl@master
-      - run: flyctl deploy --config fly.staging.toml --remote-only -e GIT_SHA=$(git rev-parse HEAD)
+      - run: flyctl deploy --config fly.staging.toml --remote-only -e GIT_SHA=$(git rev-parse HEAD) -e ORIGIN="https://flow-docs-staging.fly.dev"
```
10thfloor, line 94: Nice. This might be better than adding it via `fly secrets set`

schpet: yeah, i like this a bit better because

- it's in code
- the value will be exposed when you run `flyctl config env -c fly.staging.toml` unlike secrets which are hashes

Issue #291, opened 2022-06-22
Convert all a tags to Remix links


alse:
- Create AppLink component that uses `import { Link } from "@remix-run/react"`
- Switch all a tags to use AppLink
- Add className param. If className is not present, apply hover:opacity-75 class
- Wrap storybook layout in a MemoryRouter

PR #409, opened 2022-07-07
Convert a tags to Remix links

alse:
#291


app/ui/design-system/src/lib/Components/AppLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+        {children}
+      </a>
+    )
+
+  return (
+    <Link to={to} className={classes} title={title} style={style}>
+      {children}
+    </Link>
+  )
+}
```
schpet, line 40: random TS trivia, but an alternative type for the props on this that's might be spreadable is something like


```tsx
const AppLink = ({to, ...props}: Omit<JSX.IntrinsicElements['a'], "href"> & {to: string}) => {
  if (isLinkExternal(to)) {
    return <a href={to} {...props} />
  } else {
    return <Link to={to} {...props} />
  }
}
```

schpet (approved):
🤙🏻

app/ui/design-system/src/lib/Components/AppLink.tsx
10thfloor: `Omit` !

alse: Nice! I went with explicit props because `a` and `Link` have different types. Could also reverse it to a Pick<...similar attributes>

Issue #415, opened 2022-07-08
[Jul 8 Launch] UI Issues


srinjoyc:
# Major 

- [x] 🚨 Image & video issues [Jul 8 Launch] @10thfloor 
-  Images and videos are still missing (ie. /learn/kitty-items).
-  office hours logo is missing (resolved by #418)

- [x] 🚨 URLs [Jul 8 Launch] (@aureentuluvacal)
- When a URL ends with `/` ie. `/flow/dapp-development/flow-dapp-anatomy/` vs `/flow/dapp-development/flow-dapp-anatomy` - the first one breaks the page. This may be common while sharing and referencing links so it shouldn't break the page.

- [x] #432
- Really needed specifically for this case: `https://flow-docs-staging.fly.dev/flow/dapp-development` - in order for the user to see there are other pages here, they need to click on 'next'.

- [x] 🚨 Move the following into`/tools/xxx` route: [Jul 8 Launch] (@codingone21)
Issue Created: https://github.com/onflow/next-docs-v1/issues/417
- fcl-js
- flow-js-testing
- flow-go-sdk
- flow-cli
- emulator
- Please also search and replace the links to match this new routing

- Move `/learn/nft-marketplace` to `/flow/nft-marketplace` 
  - Include in 'Flow' section in left sidebar

- [x] 🚨  Repo Switcher Contents: [Jul 8 Launch] (@joekrill)
```
Tools
- CLI
- FCL
- Go SDK
- HTTP API
- Emulator
- VS Code Extension
- All tools (goes to /tools) 

Learn
- Cadence
- Kitty Items
- Concepts & Guides (goes to /flow)
- All content (goes to /learn)

Nodes
- Operation
- Staking
- Flow Port
- All nodes (goes to /nodes)
```

# Minor
- [x] Page Routing to subsections ⚠️ [Jul 8 Launch]: (@neilZon - 2)
  - A url like `https://flow-docs-staging.fly.dev/tools#sdks` should go directly to the SDK section. Currently it does not scroll down to the section as it should.

- [x] Secondary navigation on landing pages (@aureentuluvacal)
  - Maybe use the purple text as the selected section over the current box - similar to the navigation of internal pages. 
 
 - [x] On main nav ⚠️ [Jul 8 Launch]: (@10thfloor )
   - Show external links with the upward arrow to indicate that.
   - Main nav should close once a link is clicked. 

[...]

PR #421, opened 2022-07-08
Handle URLs with trailing slashes

aureentuluvacal:
Addresses #415 

Pages should load with or without a trailing slash. Try http://localhost:3000/flow/dapp-development/flow-dapp-anatomy/ and http://localhost:3000/flow/dapp-development/flow-dapp-anatomy. They should both load.


10thfloor (approved):

10thfloor:
I noticed this was broken on staging. 
Trailing slash hacks make me nervous, but we'll add this for now.

PR #431, opened 2022-07-11
Adds submenu on hover to landing page github links

### LINKED ISSUE #380: Add hover state to github icon on landing pages
When hovering over the github icon on the landing pages, we should display two links:
- first link should be to the github repo for the onflow github org - "All onflow repositories"
- the second one should be to the specific data for this page- "Edit this page" which should direct the user to the `component-data` for the respective page. 
[Design](https://www.figma.com/file/Bf3brjLBdiunIH8IIfcWcr/Permissionless-Deployment?node-id=199%3A869)
image

joekrill:
Adds this popover on hover to the main landing pages github link that appears fixed to the right hand side of the screen on larger screen sizes:

Screen Shot 2022-07-11 at 1 41 15 PM

If an `editPageUrl` is not provided for that page, the behavior is unchanged (no popup is shown and it acts as it did before). If it is provided, both the primary URL and the edit page url show in a popup menu. Clicking on the primary icon/link still works in either case. 

I added a delay to the hiding of the popover using lodash's `debounce` because if you hover the icon then move your mouse to the newly-revealed popover you may (and likely will) briefly move your mouse outside of one of those DOM elements, causing the popover to disappear. Which would be really annoying and frustrating. We may need to adjust the actual delay amount, however (currently it's 200ms).

Closes #380


app/ui/design-system/src/lib/Components/LandingPageLinks/GithubLinks.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const onMouseEnterHandler = useCallback(() => {
+    onMouseLeaveHandler.cancel()
+    setIsHovering(true)
+  }, [onMouseLeaveHandler])
+
+  useEffect(() => {
+    return () => {
+      onMouseLeaveHandler.cancel()
+    }
+  }, [onMouseLeaveHandler])
```
schpet, line 71: good stuff 👍🏻

10thfloor (approved):

PR #434, opened 2022-07-11
Scroll to top after navigation in internal pages

aureentuluvacal:
Addresses #415 which is the last UI fix related to the fixes @neilZon was working on.


10thfloor:
Note the `true` parameter is not supported on iOS so this just snaps the page into view on Safari mobile, instead of a nice smooth scroll. 🤷

PR #447, opened 2022-07-12
Allow trusted repos that skip rehypeSanitize

joekrill:
Re-adds the `isTrusted` property to the `ContentSpec`s that was originally introduced by #387 but accidentally reverted.

### DIFF
--- a/app/cms/compile.mdx.server.ts
+++ b/app/cms/compile.mdx.server.ts
@@ -55,11 +55,17 @@ const remarkPlugins: U.PluggableList = [
   ],
 ]
 
-const rehypePlugins = (repoName: string): U.PluggableList => {
-  return [
+const rehypePlugins = (
+  repoName: string,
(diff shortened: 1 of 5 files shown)


10thfloor (changes requested):
@joekrill  this looks good. 


Since this deals with the content mapping, we're trying to add a little more QA process into these PRs. 

Go through the following checklist and verify each route is diplaying how you expect: 


 Looks good but I think we want to introduce more caution when making changes to all things routing. 
 
 Could you verify all internal page subsections still render for important routes in each subsection? (Just click around in a few pages to make sure everything looks right, and fill the checkboxes when you're done)

- [x] `/flow/...`
- [ ] `/tools/...`
- [ ] `/cadence/language/...`
- [ ] `/cadence/tutorial/...`
- [ ] `/learn/...`

10thfloor (approved):

PR #460, opened 2022-07-17
Cp/add e2e tests

aureentuluvacal:


aureentuluvacal:
@10thfloor According to Hannah:

```
the sitemap should be dynamically generated now

we can double check by visiting the sitemap.xml on staging

https://flow-docs-staging.fly.dev/sitemap.xml
but the sitemap is generated from the sidebars. If we change the logic on how we generate sidebars (I think mack is working on generating it dynamically) then the sitemap logic might get affected
```

PR #468, opened 2022-07-19
Routing Refactor

10thfloor:
## Routing Refactor 
Separate site deployment from content deployment

#### Problem
- Overlapping patterns for managing  _route state_ across different parts of the application (prev + next buttons, breadcrumbs, sidebar, layout...etc) are making it difficult to reason about _route-state_.
- Encoding route-state in the app's code caused a fair amount of duplication. 
- We have a unique requirement that users should be able to "plug-in" content via github. This is the theme of the whole system. Call it a "distributed cms".

This site should be like a _window_ to content on GitHub. A magic mirror that you can _point at a repo_ (from a URL) and you get amazing documentation.

Continuing with the design to separate site deployment from content deployment, this PR aims to allow content to be added to the site, and integrated into the theme for end-user enjoyment, _without modification or deployment of the code in this repo_.

Assuming this is true (it is):
- All documents added to github are immediately available to our site via a unique URL.
- We can store and retrieve metadata related to any URL from GitHub.

This is incomplete but ready / in need of more technical review:

- [x] Metadata fetching for routes. 
    - Metadata that conforms to a pre-defined schema, for any route (only json is supported so far) can be loaded on demand. For example:  side-menu configuration.
    - Filesystem based: _The app does not encode the routing structure in JavaScript, so adding routes can be done usingg filestsytem templates. 
 
 - [ ] Add and integrate new in-site routes** via a PR to this repo.
    
 ### Questions?
What (if any) are the security implications of doing things this way?
---
*ok not quite, but this PR has a pattern we could expand for managing content on github.
**I'm calling them 'base' routes, but similar to "/learn", "/tools" ...etc.)


joekrill:
Is there much benefit to have the additional `__sidebar-layout` as opposed to just adding that logic to the `__document`? 

How do we want to move forward with getting these enhancements merged into main and start using them? It looks like right now this just renders some JSON and breaks breadcrumb functionality, so we probably don't want to merge this as is?

PR #480, opened 2022-07-25
Dockerize full development environment

joekrill:
This adds the entire app to the `docker-compose.yml` so that a full development environment can be brought up with a single `docker compose up`. It starts the Docs Site on port 3000 and Storybook on port 6006 to maintain the same behavior we currently have. 

This should also fix the issues some of us have been having where `yarn dev` doesn't rebuild correctly and leaves a hanging node process, necessitating restarting the app and manually killing any node instances (it doesn't use `run-p`, instead watching for CSS/SVG changes in separate docker containers).

Hoping everyone is cool with this. It should make getting things up-and-running for new folks much easier, and helps a lot when switching between other projects.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,8 @@
+FROM node:16
+
+WORKDIR /opt/next-docs
+COPY package.json yarn.lock ./
+RUN yarn
+COPY . .
+
+CMD ["yarn", "run", "dev"]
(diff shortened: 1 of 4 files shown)


10thfloor (approved):
Looks good! 
We'll have. to get around to removing postgres at some point, but this is not critical.

joekrill:
> Looks good! We'll have to get around to removing Postgres at some point, but this is not critical.

Not sure I follow... why would we remove Postgres?

10thfloor:
> > Looks good! We'll have to get around to removing Postgres at some point, but this is not critical.
> 
> Not sure I follow... why would we remove Postgres?

It's not being used for anything, but I think the client code is still part of the app which will throw an error if t doesn;t find the ocnnection string in the env.

Issue #478, opened 2022-07-25
Get breadcrumbs working in route refactor


schpet:
e.g. on a path like `/flow/concepts` or `/flow/concepts/transactions-signing` ensure that the breadcrumbs work like how it does on staging (e.g. https://flow-docs-staging.fly.dev/flow/concepts) 

start work on this from the `route-refactor` branch and not `main` – use `route-refactor` as the branch to merge PRs into

PR #481, opened 2022-07-25
Fix internal page layout

joekrill:
This fixes the internal page layout to work with the new routing refactor


schpet (approved):
nice! does this fix #478 as well?

joekrill:
> nice! does this fix #478 as well?

No that will have to be done separately.

PR #489, opened 2022-07-28
Add redirect logic.

10thfloor:


joekrill:
I wonder if we should be redirecting _all_ requests (at least "doc"/catchall requests) to strip any trailing slashes? I'm sort of doing this already "internally" in the refactor changes (that is, I'm ignoring trailing slashes) - but we're not actually _redirecting_ in those cases (so `/cadence/` and `/cadence` both render the same thing). I think it would simplify things to always redirect to a "canonical" URL (one without a trailing slash?). But I'm not sure if that would be problematic in any cases?

10thfloor:
@joekrill kill all trailing slashes!

PR #490, opened 2022-07-28
Show selector on flow page

### LINKED ISSUE #453: Enable Repo Switcher on all left menus
For routes like `/flow` and `/nodes`. The repo switcher does not show up but should.

aureentuluvacal:
Closes #453 

The switcher already showed up on the `/nodes` page so this just addresses `/flow`. I have default text that I just chose to show as an example. Happy to update it to whatever would be most user friendly.

image

### DIFF
--- a/app/routes/$repo/$.tsx
+++ b/app/routes/$repo/$.tsx
@@ -156,9 +156,9 @@ export default function RepoDocument() {
   //
   // Logically a tool is NOT included in the list of flow routes (routes with the parent /flow)
   // const tool = ![...ROUTING_STRUCTURE.flow].includes(content.contentName)
-  const isSwitchContent = Object.keys(switchContents).includes(
-    content.contentName
-  )
+  const isSwitchContent =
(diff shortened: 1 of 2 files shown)


app/routes/$repo/$.tsx
```diff
@@ -156,9 +156,9 @@ export default function RepoDocument() {
   //
   // Logically a tool is NOT included in the list of flow routes (routes with the parent /flow)
   // const tool = ![...ROUTING_STRUCTURE.flow].includes(content.contentName)
-  const isSwitchContent = Object.keys(switchContents).includes(
-    content.contentName
-  )
+  const isSwitchContent =
+    Object.keys(switchContents).includes(content.contentName) ||
+    content.contentName === "flow"
```
aureentuluvacal, line 161: Chose to put it here because if I added a "flow" key to `switchContents` it wouldn't make sense since we don't want an item to show up in the menu for `/flow`

PR #494, opened 2022-08-02
Add override

### LINKED ISSUE #454: Highlight current folder on aggregate pages (left hand menu)
For routes like `/flow` and `/nodes`. The left menu should highlight the current folder that the page being shown is (which should be the first one in the menu).

aureentuluvacal:
Closes #454 

This seems a little weird to me, but I can't come up with a more extensible way of doing this without a complete overhaul of the sidebar component. I don't think that would be worth it.

image

image

### DIFF
--- a/app/ui/design-system/src/lib/Components/InternalSidebar/index.tsx
+++ b/app/ui/design-system/src/lib/Components/InternalSidebar/index.tsx
@@ -1,4 +1,4 @@
-import { NavLink } from "@remix-run/react"
+import { NavLink, useLocation } from "@remix-run/react"
 import clsx from "clsx"
 import {
   InternalSidebarMenu,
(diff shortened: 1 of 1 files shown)


10thfloor (approved):

joekrill:
Yeah I think we should rethink what we're doing here in general. It might be best to simply redirect `/node` to `/nodes/node-operation` and redirect `/flow `/flow/dapp-development` (and TBH I thought that's what we were planning on doing here). Otherwise we will have 2 pages rendering the same exact content (but with the sidebar being different). This is going to be confusing for users, and I suspect will also be problematic for search indexing (because you'll see two results with the same content/title). 

Either that or we should have some distinct content for `/flow` and `/nodes` that acts like more of an overview page.

@10thfloor any thoughts on this? Perhaps Alex has some input (not sure of his github name, if he has one, to @-mention him).

10thfloor:
I think we're working on landing pages for these so we should redirect for now.

10thfloor (approved):

app/ui/design-system/src/lib/Components/InternalSidebar/index.tsx
```diff
@@ -60,10 +60,14 @@ export const TEMP_SIDEBAR_CONFIG: InternalSidebarConfig = {
 }
 
 export function InternalSidebar({ config, menu }: InternalSidebarProps) {
+  const location = useLocation()
+  const itemOverride =
+    location.pathname === "/flow/" || location.pathname === "/nodes/"
```
10thfloor, line 65: Hate to see this, but seems unavoidable with the current setup. Going tohold of on merging this until @joekrill finishes up with the routing changes.

joekrill:
I don't _think_ this will be necessary anymore - but i'm not 100% sure. In the refactor changes each collection can specify a `redirects` section which is just a mapping of paths to redirect (i.e. 016c3486eefa795cb9ba51025b90f486997c870e4ca7e2106377e11dc070R100). I didn't add these particular redirects for some reason, but maybe we should handle that there instead?

joekrill:
Closing this because I believe this has been implemented in other ways since the routing refactor.

Issue #492, opened 2022-07-29
Design Audit Fixes


10thfloor:
### Please address each fix in a self-contained commit! 

We've done an initial design audit, and the results are in. There are a bunch of minor fixes, listed here: 
https://www.notion.so/turtle/Design-QA-c4253f06293f47daa22229acebed1d13

As well as a visual reference here: 
https://www.figma.com/file/hZOowo88kSUlDL1D8fbyGY/Flow-Developer-Hub-QA

### Items to address 

Here's a list of actionable items to address:
- [x] 4	Home	Smart Contracts block icon missing line @aureentuluvacal 
- [x] 5	Home	Missing Submit a tool button @aureentuluvacal 
- [x] 8	Home	Upcoming events small card carousel is cut off @aureentuluvacal 
- [x] 9	Home	Missing subscribe to newsletter @aureentuluvacal 
- [x] 11	Home	Size of hero header type too small on mobile @aureentuluvacal 
- [x] 12	Home	Missing spacing on mobile between start your project & quickstarts section @aureentuluvacal 
- [x] 13	Home	Upcoming Events cards image being awkwardly sized @aureentuluvacal 
- [x] 14	Home	Upcoming events Slider buttons way too big and using green as highlight colour @aureentuluvacal 
- [x] 15	Nav	        Dark/light mode toggle showing on mobile and the word “Search” @aureentuluvacal 
- [x] 16	Nav	        Mobile nav incomplete, missing hierarchy on the main nav and status, subnav categories have been added and once in the sub-sub-nav pages; cards are off and no icons  @aureentuluvacal
- [x] 17	Search	Search missing bottom hotkey instructions & “View X Results” @alse
- [x] 19	Interior	Missing gradient on interior nav @alse
- [x] 20	Interior	Edit Github button should be blue @alse
- [x] 21	Interior	Edit on Github button should be hidden on Mobile @alse
- [x] 22	Interior	Type hierarchy on docs content should have bold main Header and Subheaders in the content, get rid of line below subheaders, add more spacing between paragraph and subheader. @alse
                * _The line below subheaders comes from the markdown content. We'd have to either hide hr tags, or update all_ the docs
- [x] 24	Interior	Lack of author, date, etc. on interior docs @alse
- [ ] 25	Interior	On Mobile, different nav pattern. “on this page” has a different design and sits above the title. main navigation has been added to the breadcrumbs @aureentuluvacal 
[...]

PR #497, opened 2022-08-03
Add search dialog shortcuts and misc search improvements

alse:
#492 `17 Search Search missing bottom hotkey instructions & “View X Results”`

* "View X Results" is still missing because we don't have a search results page yet

image
image


10thfloor (approved):
Looks great.

Issue #501, opened 2022-08-08
Cache assets


10thfloor:
Some assets in some docs are stored in Github, and the GitHub API is used to fetch them. 
We need to cache these GitHub hosted images to avoid going over the GitHub API request rate limits: 

### Possible Strategies:
- Cache using REDIS
- HTTP Headers TTL 
- Cache using ServiceWorker

PR #504, opened 2022-08-08
Initial routing refactor and reorg

joekrill:
This should (hopefully) fix a lot of the routing issues we've been having and hopefully simplify the logic around URLs. Apologies for the large PR - it was difficult to refactor this in smaller pieces due to how the previous implementation worked. The main points are:

- Instead of having a "repo" as a source, which can in some cases contain multiple "content sections", data now organized into "doc collections" (open to better naming) which are keyed on their root base path. Each doc collection has a "source" which is the github repo (owner/name/branch) plus `rootPath` where the content lives. Multiple doc collection can point to the same repo, but would generally have different `rootPath`s.

- Each "doc collection" also has a "manifest" property which specifies the sidebars, display names, etc. This should allows us to more easily pull these configurations out into separate files or put them in the underlying source repo itself.

- Sidebars definitions are also keyed by _their_ base path (which is relative to the collection base path). A doc collection can specify sidebars an any location. In the future we could allow content create to put something like a `sidebar.json` in any arbitrary directory. Sidebar URLs should be specified relative to their base path. 

- We can't use Remix's routing because relative `Link`s in react-router (and therefore Remix) are relative to the nearest `Route` that it's rendered it. But "splats" are not included in that. So we can't support arbitrary-depths of content that way (we may be able to if Remix would let us render `<Route />` elements, but that doesn't work). So instead the catch-all route (`/_docs/_layout/$.tsx`) passes the base URL into the React tree using `InternalUrlContext`. The `InternalContentLink` can then prepend the appropriate path. This eliminates the need to rewrite the links using a rehype plugin. 

- Similar logic is used for "assets" (images, currently) - but instead of prepending just the base path, we also prepend `/assets` so we can send those requests directly to the correct  endpoint (previously we were redirecting if we thought the request looked like an image request. 

Some additional small changes:

- docs: Adds "Useful docker commands and tips" to README
- docker: Add "app-yarn" to run on startup to make sure node_modules is up-to-date
- app: adds a logging service shell to centralize logging calls (currently just maps to console.log calls)
- Renames props for consistency: prefer "title"/"href" prop names for links, etc.
- sidebars can be arbitrarily deep (though at some point visually that might not be practical)

This isn't 100% complete, but given we'd like to launch shortly I think it's "good enough" and provides a lot of improvements. There are some small issues that we need to resolve before Wednesday:

- We need to make sure the content itself uses relative links. I've noticed this does not seem to be the case everywhere (cadence, for example)
- There's some issues with "index" routes and navigating to and away from them that need to be fixed. 
- Redirects (`manifest.redirects`) are not working with client-side navigation. 

And some other less-pressing issues:

- Links in the `headers` section of the manifest should also be specified as relative to their base paths so they are consistent with how sidebars work (and then we could ultimately push those to the content repos, too). (We only have two of these currently so it's manageable to keep them relative to the "collection root" for now. 
- `app/routes/action/refresh` needs to be fixed. But it wasn't entirely clear how this should work - we need to regroup on the caching strategy in general, and how we decide what needs to be removed from the cache when we "refresh".
- We should add some additional logging (server-side).
- Assets (`/assets/*`) should be cached when fetched from github


10thfloor:
Awesome. 

To quickly address some of the TODOs: 

> We need to make sure the content itself uses relative links. I've noticed this does not seem to be the case everywhere (cadence, for example)

I'm working on a "content authoring guide" which will cover the constraints, such as this, and other importnat gotchas wen writing docs for display on the site, (eg, have only 1 file per page, prefer .md extension...etc) that should cover this .

> app/routes/action/refresh needs to be fixed. But it wasn't entirely clear how this should work - we need to regroup on the caching strategy in general, and how we decide what needs to be removed from the cache when we "refresh".

The refresh endpoint receives a list of changed files from any PR to any repo we're rendering, via a github action that is deployed to each repo. It should remove any files in the list from the cache, and then use the existing fetching logic for pages to re-request and cache the new pages. It was working at one point, but yes this needs to be re-tested.

> We should add some additional logging (server-side).
We have an issue for this now! https://github.com/onflow/next-docs-v1/issues/501

> Assets (/assets/*) should be cached when fetched from github
And this too! https://github.com/onflow/next-docs-v1/issues/501

10thfloor (approved):

joekrill:
> I'm working on a "content authoring guide" which will cover the constraints, such as this, and other important gotchas when writing docs for display on the site, (eg, have only 1 file per page, prefer .md extension...etc) that should cover this.

Awesome! I wanted to make sure it was noted here because if someone pulls this branch to test it, some things won't be working because of this.

> The refresh endpoint receives a list of changed files from any PR to any repo we're rendering, via a GitHub action that is deployed to each repo. It should remove any files in the list from the cache, and then use the existing fetching logic for pages to re-request and cache the new pages. This accounts for updated pages, as well as any pages that were added or removed. It was working at one point, but yes this needs to be re-tested.

Yeah I think we may still have some incorrect behavior around index pages. For example we cache based on the _request_. So for index pages, for example, the request for both `/` and `/index` render the same page, and may be getting cached twice. I'm not 100% sure about this, I haven't looked into it in enough detail.

PR #506, opened 2022-08-09
feat: add wallets section to tools page

alxflw:
Added a "Wallets" section to the tools page:

Screen Shot 2022-08-08 at 10 14 03 PM

Added a "Flow Wallets" link to the nav:

Screen Shot 2022-08-08 at 10 19 42 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ git clone https://github.com/onflow/next-docs-v1.git
 You'll need to acquire the project's `.env` file before continuing.
 
 1. Add the `.env` file to the project root.
-2. `docker compose up`
+2. `yarn build`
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -5489,6 +5489,14 @@ axios@^0.21.1:
   dependencies:
     follow-redirects "^1.14.0"
 
+axios@^0.27.2:
```
joekrill, line 5492: Are these yarn.lock changes intentional? I think we should only see changes here if something was upgraded or if dependencies were changed in `package.json`?

joekrill (approved):

10thfloor (approved):

PR #530, opened 2022-08-10
Adjust mobile navbar

aureentuluvacal:
Addresses #492 

Reformatted mobile navbar, removes the intermediate tab "page" in favor of directly rendering the section.

image

image


app/routes/poll-network.tsx
```diff
@@ -16,9 +16,14 @@ export async function loader(): Promise<LoaderData> {
   }
 }
 
+const LS_KEY = "flow_network_data"
+
 export default function () {
   const networkData = useLoaderData<LoaderData>()
-  const [status, setStatus] = useState(networkData)
+  const cachedStatus = localStorage.getItem(LS_KEY)
```
joekrill, line 23: This won't work if rendered server-side, will it? I don't think `localStorage` will be available on a server render. https://remix.run/docs/en/v1/guides/constraints#rendering-with-browser-only-apis

app/routes/poll-network.tsx
```diff
@@ -40,28 +45,47 @@ export default function () {
   useEffect(() => {
     if (fetcher.data) {
       setStatus(fetcher.data)
+      localStorage.setItem(LS_KEY, JSON.stringify(status))
```
joekrill, line 48: I think this is going to cause an outdated version (the previous value) of status in `localStorage`, won't it? Should this be `JSON.stringify(fetcher.data)`?

app/ui/design-system/src/lib/Components/NavigationBar/MobileMenu.tsx
```diff
@@ -1,39 +1,54 @@
...
 
 export type MobileMenuProps = {
   menuItems: MenuItem[]
 }
 
 export function MobileMenu({ menuItems }: MobileMenuProps) {
-  const [selectedIndex, setSelectedIndex] = useState(-1)
-  const selectedMenuItem = menuItems[selectedIndex] as
-    | DropdownMenuItem
+  const [selectedIndex, setSelectedIndex] = useState<[number, number]>([-1, -1])
```
joekrill, line 12: This is a little confusing to me. Why do we need two indexes and what is the difference between them? Can we make this more clear maybe?

joekrill (changes requested):

app/routes/poll-network.tsx
aureentuluvacal: I'll just take this out for now. Keep forgetting remix does SSR.

PR #544, opened 2022-08-11
Integrate dynamic open-graph image service.

10thfloor:
Generates a custom open-graph image when a page link is shared, based on the title and description form the yml frontmatter, using the designs from Turtle.

We might think of alternative sources for these as some pages may not have that in their frontmatter.
- This PR includes defaults when no title or description is found. 

Here is a preview:

Screen Shot 2022-08-10 at 6 24 52 PM

### DIFF
--- a/app/routes/__docs/$/$.tsx
+++ b/app/routes/__docs/$/$.tsx
@@ -71,11 +71,14 @@ export const loader: LoaderFunction = async ({ params, request }) => {
 
 export const meta: MetaFunction = ({ data, location }) => {
   const typedData = data as LoaderData
+  const title = typedData.page.frontmatter?.title
+  const description = typedData.page.frontmatter?.description || ""
   if (typedData && typedData.page) {
     return getSocialMetas({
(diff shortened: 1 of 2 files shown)


app/routes/__docs/$/$.tsx
```diff
@@ -71,11 +71,14 @@ export const loader: LoaderFunction = async ({ params, request }) => {
 
 export const meta: MetaFunction = ({ data, location }) => {
   const typedData = data as LoaderData
+  const title = typedData.page.frontmatter?.title
+  const description = typedData.page.frontmatter?.description || ""
   if (typedData && typedData.page) {
```
gpavlov2016, line 76: If typedData.page is undefined there would be an exception on line 74

gpavlov2016 (approved):

app/routes/__docs/$/$.tsx
10thfloor: Ok will address this.

Issue #524, opened 2022-08-10
Finish Network Page


srinjoyc:
Few remaining items to get the network page completed.

On /network:
1)  - [x] Update this information on /network (check with protocol team/comms for dates)
<img width="620" alt="image" src="https://user-images.githubusercontent.com/8709330/183917041-c2d107f7-41d9-4052-8395-314866b19a33.png">
2)  - [x] Any "node operator callout" should be the following: "Node Operators" -> Body: "View the current nodes in the network and who is running them. You can use Flow Port to delegate or stake to a node. The picture should be the one of the 'abstract images' used across the site (use best discretion)

When clicking into a details page of a network (/network/:name):
3) - [x] We need to update 
<img width="590" alt="image" src="https://user-images.githubusercontent.com/8709330/183917744-0acfce33-9b19-437b-83c1-84cf463d8194.png">

The RSS feed should point to the RSS subscription feed available at status.onflow.org (in the subscribe for updates). 

Upcoming spork date should be filled in as per protocol team.

4) - [x] For past spork information, they need to be parsed in through this json file: https://github.com/onflow/flow/blob/master/sporks.json

5) - [x] Once 4 is done, the navigation under the network tab should change the link of upcoming and past spork information to those pages. We should also remove or redirect the internal pages in /nodes that contain this same information.

aureentuluvacal:
Just gonna recycle this issue since it still captures all the pieces we need.

PR #550, opened 2022-08-11
Fetch spork data

aureentuluvacal:
Addresses #4 in #524

### DIFF
--- a/app/cms/utils/fetch-sporks.tsx
+++ b/app/cms/utils/fetch-sporks.tsx
@@ -1,4 +1,4 @@
-import { Octokit } from "@octokit/rest"
+import { octokit } from "~/cms/github.server"
 import { cachified } from "../cache.server"
 import { redisCache } from "../redis.server"
 import { SporksCardProps } from "~/ui/design-system/src/lib/Components/SporksCard"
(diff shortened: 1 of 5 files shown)


app/ui/design-system/src/lib/Components/SporksCard/index.tsx
```diff
@@ -8,7 +8,7 @@ import { SporkMetadata } from "../../interfaces"
 
 export type SporksCardProps = {
   heading: string
-  timestamp: Date
+  timestamp: string
```
aureentuluvacal, line 11: Had to change this since I was getting an invalid date error from `date-fns`.

app/cms/utils/fetch-sporks.tsx
```diff
@@ -0,0 +1,63 @@
...
+    cache: redisCache,
+    maxAge: 1000 * 60 * 60 * 4,
+    forceFresh: process.env.FORCE_REFRESH === "true",
+    key: `sporks`,
+    getFreshValue: fetchFreshSporks,
+  })
+}
+
+export const fetchFreshSporks = async () => {
+  const octokit = new Octokit({
```
joekrill, line 17: Should we just import the `octokit` instance from `app/cms/github.server.ts` instead? That way we get the throttling options and any logging pre-configured.

aureentuluvacal: That's a good call! I'll swap it out.

aureentuluvacal: @joekrill updated!

PR #557, opened 2022-08-11
Reset/restore scroll on navigation change

joekrill:
Remix does not have support for reseting/restoring scroll position anywhere except in the main window. We use a flexbox layout with an inner scrolling container, so that doesn't work for us. This change applies scroll reset/restore to our main scroll container.


aureentuluvacal (approved):
Tested and it works!

10thfloor (approved):

PR #559, opened 2022-08-11
Add overflow-auto to desktop menu

alse:
image
image


10thfloor (changes requested):
Is this what we want? I don;t think scrolling this menu is a good idea.

PR #568, opened 2022-08-12
Use explicit transparent gradient color

joekrill:
Older versions of safari do not handle transparent gradients correctly, as seen here:

Screen Shot 2022-08-12 at 9 38 36 AM

See: https://css-tricks.com/thing-know-gradients-transparent-black/


aureentuluvacal (approved):
Perfect, tested on Browserstack in an older version of Safari. Looks good!

10thfloor (approved):

PR #579, opened 2022-08-16
Migrate to Express to access Cache-Control headers

10thfloor:
Reference issue: https://github.com/onflow/next-docs-v1/issues/501
- Cache static assets in browser for longer.
- Use express so we can modify  headers.


app/routes/__docs/assets.$.ts
```diff
@@ -51,7 +51,7 @@ export const loader: LoaderFunction = async ({ params }) => {
       status: 200,
       headers: {
         "Content-Type": mimeType,
-        "Cache-Control": "max-age=300",
+        "Cache-Control": "max-age=3153600",
```
schpet, line 54: with this setup, if a repo updates any asset (e.g. docs/diagram.png), you'll get stale data for a year if they update it

10thfloor: Yeah, I'll try to narrow it to fonts only.

joekrill:
Are we sure the express stuff is necessary? I see our headings being set as expected already. I even tried setting a random header ("X-foobar") and see it being returned: 

Screen Shot 2022-08-16 at 4 29 22 PM

app/routes/__docs/assets.$.ts
schpet: fwiw the `cachified` thing that uses redis would be really helpful to cache these assets though, i think! 

we _could_ get a super long lived max-age for these if we put the git sha for those images in the url, but that would involve tweaking the params for this component and this thing:

https://github.com/onflow/next-docs-v1/blob/9b7906168f9487d1316aa612d73d5a68b1237edb/app/ui/design-system/src/lib/Components/InternalImg/InternalImg.tsx#L14

10thfloor: Yes, I've been thinking we could also use REDIS to store optimized versions of any images in external docs as well! 

Procees MD -> Find image URL -> fetch it -> convert to webp -> store in REDIS -> modify img src to point to resource route for fetching webp data from REDIS -> load return page with all img src's pointing to webp optimized versions !

10thfloor:
> Are we sure the express stuff is necessary? I see our headings being set as expected already. I even tried setting a random header ("X-foobar") and see it being returned:
> 
> Screen Shot 2022-08-16 at 4 29 22 PM

Where are you setting those headers?

app/routes/__docs/assets.$.ts
schpet: if possible i would probably recommend using an image cdn for stuff like webp conversion, e.g. imgix, cloudinary, or self hosting something like imgproxy. 

this is nice because these servers/services are easy to setup, have features like letting you serve smaller images to small screens, not fall over when a large image shows up, etc

PR #599, opened 2022-08-18
Redirect hrefs that have redirect defined in redirects.js

10thfloor:


app/cms/utils/return-redirect-for-route.ts
```diff
@@ -0,0 +1,13 @@
+import redirects from "../../redirects.js"
+
+export function returnRedirectForRoute(path: string): string | undefined {
```
10thfloor, line 3:
```suggestion
export function returnRedirectForRoute(path: string): string {
```

PR #612, opened 2022-08-23
Hide content, add manifest prop to hide doc

aureentuluvacal:
Todo:

- [x] Create a custom rehype plugin to strip out hidden content
- [ ] Add logic for generating our own ToC to avoid parsing multiple times


joekrill:
Handled by #623

PR #617, opened 2022-08-24
Refactor/consolidate component data

### LINKED ISSUE #586: Refactor component data to be consistent across all landing page components
We should have a single source of truth for all static data being loaded into landing pages and a consistent easy way to update vs having various folders and different ways of rendering the data.

joekrill:
Moves _most_ static data into the `apps/data` folder. There's still some data in `app/cms/utils/constants.tsx` (and probably other places) that should likely be consolidated here, too. But this PR was already getting quite large.

We really shouldn't be "reaching out" from the design system components into the main app -- those components are really meant to be self-contained and dependencies of the main app, and not the other way around (or both). So this refactors some of those cases and passes them down as props instead.

Also:

- Fixes an issue with postcss exiting without generating CSS when using `docker-compose up`
- Extracts each doc collection into it's own JSON file.
- Removes some unused files

Fixes #586


app/ui/design-system/src/lib/Components/Footer/index.tsx
```diff
@@ -171,7 +170,7 @@ export const Footer = ({ sections = footerSections }) => {
             <AppLink to="https://github.com/onflow">
               <img src={GithubIcon} height={32} width={32} />
             </AppLink>
-            <AppLink to={DISCORD_URL}>
+            <AppLink to="https://onflow.org/discord">
```
10thfloor, line 173: Since these are used in a couple of places, I think they should be "constants".

app/data/metadata.ts
```diff
@@ -0,0 +1,12 @@
+export const metadata = {
+  siteTitle: "Welcome to the Flow Developer Portal",
+  siteDescription: "Your resource for everything you'll need to build in Flow",
+  defaultPageTitle: "Flow Developer Portal",
```
10thfloor, line 4: Can we update this to "Welcome"?

app/data/metadata.ts
```diff
@@ -0,0 +1,12 @@
+export const metadata = {
+  siteTitle: "Welcome to the Flow Developer Portal",
```
10thfloor, line 2: Can we update this to "Flow Developer Portal"

10thfloor (changes requested):
This looks good. A couple small changes and we can merge!

10thfloor (approved):

PR #619, opened 2022-08-25
Fix network details page

joekrill:
The network details page tab navigation wasn't working correctly after initial server-side page load. The "networkStatus"  data was not being updated properly. This is because the `tabs` we were passing into the `TabMenu` included `link` properties, which never triggers the `onTabChange` callback - it just navigates to the page. 

This fixes that and simplifies a lot of the logic by adding a predefined array of networks (`app/data/networks.ts`) rather than relying on the statuspage.io response to extract the data based on the title.


aureentuluvacal (approved):
Lovely!

10thfloor (approved):

PR #623, opened 2022-08-26
Exclude docs content

### LINKED ISSUE #589: Allow authors to hide content
In some cases authors might want to thide content on the docs site, while keeping content visible in github

joekrill:
This primarily allows us to exclude content within a document from being rendered on the docs site by wrapping it in an element with the class name "docs-site-omit". i.e.:

```mdx
# This will be shown

<div className="docs-site-omit">
This will be removed entirely when rendering on the developer portal
</div>
```

But the way we generated the table of contents (ToC) was no longer usable - we were previously generating it by looking only at the MDX source. But the content we want to hide gets removed after the MDX is converted to an abstract syntax tree (AST) for processing by unified. So if there were any headers within in the hidden/removed content they would still be included in the ToC even though they were removed from the final rendered document. 

There were also some other problems with the way we were generating the ToC: We couldn't guarantee that it generated a set of unique "slugs" because if there were duplicate header titles they would generate the same slug/hash. We were also generating the slugs twice - once on the server to generate the ToC, and again on the client to generate the anchor, which left us open to possible hash differences.

So this changes how the TOC is generated. A custom rehype plugin generates a `toc` object in a way that guarantees all entries have unique slugs. It also adds a `data-slug` attribute to the element which we can then use during rendering to ensure the hash/slug will always match the one used by the ToC.

Additional small changes:
- Extracted each of our custom rehype plugins into their own files for clarity.
- Adds a `replaceNonStandardReactAttributes` plugin that handles cases where doc authors are using HTML attribute names instead of their JSX equivalents ([i.e. using `class` instead of `className`](https://raw.githubusercontent.com/onflow/fcl-js/master/docs/reference/api.md)). This is mainly needed because some files are still regular markdown (not MDX) which allows HTML, not JSX.
- Removes the `isTrusted` property/variable that was controlling which rehype plugins get applied -- this was always `true` so had no effect.
- In docker-compose, when bringing up services, waits for `yarn` to finish running before starting the dev server to ensure packages are up-to-date and prevent "module not found" errors.

Note: I did not implement the functionality that allows hiding/excluding entire files (which I believe was a misunderstanding of this issue). We should create a new issue if we want to handle those cases. 

Closes #589


10thfloor:
> Adds a replaceNonStandardReactAttributes plugin that handles cases where doc authors are using HTML attribute names instead of their JSX equivalents ([i.e. using class instead of className](https://raw.githubusercontent.com/onflow/fcl-js/master/docs/reference/api.md)). This is mainly needed because some files are still regular markdown (not MDX) which allows HTML, not JSX.

Awesome, I wanted to add this.

10thfloor (approved):

10thfloor (approved):

10thfloor (approved):

Issue #632, opened 2022-09-01
Bad path input to `/action/refresh`


10thfloor:
**Describe the bug**
Sentry is capturing exceptions from the `/action/refresh` endpoint.
Take a look at the following screenshot: 
<img width="1040" alt="Screen Shot 2022-09-01 at 2 37 18 PM" src="https://user-images.githubusercontent.com/901466/188017532-8a99b4b0-57d8-4cca-9130-0cf48be7917f.png">

- The refresh process begins at `Refreshing docs/content/kitty-items/index.md`
- In the subsequent step, an attempt is made to get the MDX for the path, (https://github.com/onflow/developer-portal/blob/main/app/routes/action/refresh.tsx#L93) But the path is not correct (`docs/content` is duplicated and the path is appended in the log and contains 2 slashes `//`)
- This results in an error and an exception captured in the last lines of the log above (you can ignore the 'searching for redirect').

**Expected behavior**
`getMdxPage` is called with the appropriate path to the content, and no error is thrown.

joekrill:
This endpoint is deprecated anyway and repos need to be updated to use the github-webhook endpoint instead.

PR #636, opened 2022-09-02
`contributions` is not always present for /refresh endpoint

joekrill:
It looks like the error in #632 is actually caused by assuming the `contributions` property will be provided to the endpoint, which isn't happening. This guards for that.


10thfloor (approved):
Will merge this for now, but will be removing this.

PR #640, opened 2022-09-06
Use full path for internal links

### LINKED ISSUE #633: Links to headings in internal pages not parsed correctly.
**Describe the bug**
Visit this page: https://developers.flow.com/learn/concepts/transaction-signing
Look for this line:  Proposer: the account that specifies a **proposal key**.

"proposal key" is a link to a heading further down the page. 

- Notice the link is also in the sidebar. 

The sidebar link is correct, and the link as defined in GitHub is also correct. You can see the link is defined as `[proposal key](#proposal-key)` (Source: https://github.com/onflow/flow/edit/master/docs/content/concepts/transaction-signing.md)

**Bug**:
The docs site creates the following link after the page has been processed:
` https://developers.flow.com/learn/concepts/#proposal-key`

- The link has an unnecessary slash after `/concepts`
- The link is also missing the parent page segment.

This is causing exceptions in the app.

The correct link should be `/learn/concepts/transaction-signing#proposal-key`

**Expected behavior**
links in the sidebar and the links on the page, that link to the same internal heading, should be the same.

aureentuluvacal:
Closes #633

### DIFF
--- a/app/cms/utils/strip-slashes.test.ts
+++ b/app/cms/utils/strip-slashes.test.ts
@@ -2,7 +2,7 @@ import { expect, test, describe } from "vitest"
 import {
   stripSlahes,
   stripStartingSlahes,
-  stripTrailingSlahes,
+  stripTrailingSlashes,
 } from "./strip-slashes"
 
 describe("stripSlahes", () => {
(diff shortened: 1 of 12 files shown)


app/cms/utils/strip-slashes.test.ts
```diff
@@ -2,7 +2,7 @@ import { expect, test, describe } from "vitest"
 import {
   stripSlahes,
   stripStartingSlahes,
-  stripTrailingSlahes,
+  stripTrailingSlashes,
```
aureentuluvacal, line 5: Typo fix

app/routes/__docs/$/$.tsx
```diff
@@ -58,7 +59,10 @@ export const loader: LoaderFunction = async ({ params, request }) => {
       data.collectionRootPath,
       page.origin.relativePath
     )
-    const pageBasePath = nodePath.posix.dirname(path) + "/"
+    const pageBasePath =
```
joekrill, line 62: Maybe we should rename this? It's not really a "base path" anymore -- or at least it's more confusing now. Because it's actually the path to the full page now. 

Is there even a need to extract the `dirname` and `basename` and piece them back together? I think this whole thing is just equivalent to `const pageBasePath = stripExtension(path)` now, isn't it?

10thfloor: This makes sense to me. 
The only thing I'm unclear about is why were we separating things in the first place?

10thfloor: So many possibilities. lol.

PR #647, opened 2022-09-08
Do not use full URL when determining docs canonical URLs

joekrill:
Use the extracted `path` and not the full `request.url` when setting the canonical URL for docs pages. 

When using TLS, it is terminated before it hits the app (load balancer or gateway), so from the application's perspective the request is just a plain `http` request. So for canonical URLs we shouldn't use `request.url` as-is because it may have the wrong protocol (or even origin, for that matter). Just passing the `path` we extracted will use the `ORIGIN` as a base and generate a URL based on the correct protocol/origin.


schpet (approved):
👍🏻

PR #660, opened 2022-09-16
Preview docs from different branches

### LINKED ISSUE #654: View a preview link on staging to preview my changes
## Feature Story 
As a contributor making changes to the document site, I would like to view preview links for my changes so I can ensure they look as expected. 

## Notes 
Get preview links for changes posting in github comments. 

[Product Requirements Document](https://docs.google.com/document/d/1Au9A0eEGJE_VhVYBcAKVUdwkJn2SsGOVfj2u-ckcTHY/edit?usp=sharing)

## Technical Task List: 
- [x] Enable a preview mode that can be toggled on via environment variable
- [x]  __docs routes should handle requests that include a ?preview=<sha>
  - [x] Internal links within the rendered page should also include the `preview` query string
  - [x] Sidebar links should include `preview` query string
  - [x] next/prev links should include `preview` string`
  - [x] Metadata heading should use preview branch data
  - [x] Show an error when the ref/branch is not valid/found
- [x]  Do not generate canonical <link /> when in preview mode
- [x] Add robots.txt rule to ignore paths with preview query string
- [x]  Include <meta name="robots" content="noindex"> in rendered pages
- [x]  Update InternalUrlContext and InternalSidebarUrlContext
- [x]  Assets.$.ts should consider `preview` parameter
- [x] Add a 'Preview' label somewhere on the page to annotate you are viewing this in preview

joekrill:
Closes #654 

This allows specifying a `preview=` query string parameter which takes a ref or branch name and will fetch docs from that branch/ref instead of the one specified by the doc collection manifest.

It looks something like this:

Screen Shot 2022-09-16 at 11 40 25 AM

All relative URLs within that document collection also include the `preview=` parameter: sidebar links, "internal" doc links, next/previous links, asset URLs. 

If the branch does not exist and error indicates that:
 
Screen Shot 2022-09-16 at 11 40 34 AM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ apps/**/.cache
 .vscode
 
 node_modules
+package-lock.json
 
 /.cache
 /build
--- a/app/cms/collections.server.ts
+++ b/app/cms/collections.server.ts
@@ -1,12 +1,13 @@
 import Ajv from "ajv"
+import { ensure as ensureError } from "errorish"
+import { RequestError } from "@octokit/request-error"
 import { posix } from "node:path"
 import invariant from "tiny-invariant"
 import { cachified, downloadFileByPath, redisCache } from "~/cms"
 import manifestSchema from "~/data/doc-collection-manifest-schema.json"
 import { InternalLandingHeaderProps } from "~/ui/design-system/src/lib/Components/InternalLandingHeader"
 import { SidebarItemList } from "~/ui/design-system/src/lib/Components/InternalSidebar"
-import { findMostSpecificPath } from "./utils/find-most-specific-path"
-import { stripSlahes } from "./utils/strip-slashes"
+import { docCollections } from "../data/doc-collections"
 import {
   DocCollection,
   DocCollectionManifest,
@@ -14,7 +15,8 @@ import {
   JSON_MANIFEST_FILENAME,
   manifestCacheKey,
 } from "./doc-collections.server"
-import { docCollections } from "../data/doc-collections"
+import { findMostSpecificPath } from "./utils/find-most-specific-path"
+import { stripSlahes } from "./utils/strip-slashes"
 
 export const collectionPaths = Object.keys(docCollections)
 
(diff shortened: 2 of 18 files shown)


10thfloor (changes requested):
Nice to see this working. Could we adjust the banner colour to something more neutral, maybe light blue or green? Red feels scary. 😅

joekrill:
sure - how's cyan? 

Screen Shot 2022-09-16 at 3 40 56 PM

10thfloor:
@joekrill this is great. Are you planning another PR that posts links to new docs in Pull Request comments?

10thfloor (changes requested):
Cyan is nice. Omit the hash from the preview banner as well. The banner should have a _link_ to the in-progress document on Github.

The text should read: "You are viewing an unpublished preview. View page in GitHub."
And the "View page in Github" should be a link, with the same icon for external links we use elsewhere.

joekrill:
Sorry @10thfloor I guess I accidentally edited _your_ reply instead of _adding_ my reply? I have no idea how that happened, that was totally unintentional.

10thfloor:
> > @joekrill I was also thinking this over more and was wondering why we decided to use the commit hash?
> 
> Actually any ref will work - commit hash, branch name, tag, etc...
> 
> > Can we use the PR number instead of a sha? so the query string would be ?pr=420 instead of preview=sha ?
> 
> It is _possible_, But:
> 
> 1. We'd need to make an extra API call to Github on every request to determine the head ref of that PR.
> 2. It may not be possible to determine the PR at the time we receive the webhook in which we generate the preview URLs automatically. That webhook gets sent when a _branch_ is changed, and that may have zero, one, or more PRs associated with it.
> 
> > I think this is much better from an end-user perspective. Finding the PR number is more straightforward than the commit hash.
> 
> I suggest that we generate preview URLs using branch names when possible (with fallback to SHA), for now, and if we want to use PR numbers we implement that as an enhancement.

Ok, yes branch names are also user friendly, and I think better than the sha and having both means to access a preview is great.

10thfloor (approved):

PR #696, opened 2022-10-04
Refactor reusable code from webhook server.

10thfloor:


app/cms/github-webhook.server.ts
```diff
@@ -14,9 +14,11 @@ type ProcessResult = {
...
 }
 
-export function pushEventCacheKeysToInvalidate(
-  event: PushEvent
-): ProcessResult {
+export type PathsResult = {
+  updatedDocuments: Set<string>
+}
+
+function getMatchingDocCollectionsForPushEvent(event: PushEvent) {
```
joekrill, line 21: There's a `getCollectionsForRepo` at `app/cms/doc-collections/get-collections-for-repo.ts`... perhaps we could use that instead here? Just modify it to take an option branch parameter?

app/cms/github-webhook.server.ts
```diff
@@ -32,6 +34,70 @@ export function pushEventCacheKeysToInvalidate(
...
+  ])
+}
+
+function getPathsOfDocumentsFromChanges(
+  docCollection: any,
+  allChangedFiles: string[]
+): string[] {
+  return allChangedFiles.filter((path) => {
+    let isDocumentPath = path.startsWith(docCollection.source.rootPath)
+    let isDocument = path.endsWith(".md") || path.endsWith(".mdx")
```
joekrill, line 58: There's a `filterFileNameHasMarkdownExtension` at `app/cms/github/filter-file-name-has-markdown-extension.ts` - might be better to use that here? (Though looking at it, the regex should probably be modified with the "i" option, i.e. ` /(.mdx?)/i`, to ignore case).

10thfloor: This code and the above example were already written in this file, I just moved them into functions.

PR #781, opened 2022-12-08
Add test framework for content checker + limit it to only look at modified files

huyndo:
- [x] Add a unit test framework that mocks Github APIs and allow local development of the content checker.
- [x] Add documentation for this framework
- [x] Limits content checker to only look at modified files from the PR


bthaile (approved):
great! really like this readme
code makes sense,

PR #790, opened 2022-12-13
Add content checker workflow to main

huyndo:


bthaile (approved):
lgtm, need to merge to main as well as production

PR #813, opened 2023-01-19
aded hotjar init to dev portal

MrDSGC:
- added hotjar package to dev portal
- added env variable for hotjar site id


bthaile (approved):
lgtm

PR #814, opened 2023-01-19
Add asset ignore check

bthaile:
Need to ignore assets url in order to remove false positive for asset url validation


app/cms/doc-validation/test/url-verification.test.ts
```diff
@@ -86,6 +98,7 @@ it("Test urls", async () => {
   // TODO: spin up mock http server to test validating urls.
   testUrls.forEach(async (item) => {
     const urlResult = await validateUrl(item.urlItem as UrlItem, context)
+    console.log("testing urls", urlResult)
```
MrDSGC, line 101: delete log

MrDSGC (approved):
console log then lgtm

PR #818, opened 2023-01-20
fixed hotjar integration

MrDSGC:
- this works but site id is exposed.
- process is undefined in `entrypoint.client.tsx`


bthaile (approved):
temp solution for getting hotjar configured

PR #828, opened 2023-01-23
remove markdown comments, they are not converted to html

bthaile:
Remove markdown comments, they are not converted to html and content checker blows up

closes: https://github.com/onflow/developer-portal/issues/827


justinbarry (approved):
LGTM

PR #837, opened 2023-01-24
update version as suggested in sentry monitoring UI

bthaile:
update version as suggested in sentry monitoring UI


fotescodev (approved):
Looks good!

PR #847, opened 2023-01-26
removed kitty items from dropdown and replaced with Flow Mobile

MrDSGC:
- removed kitty items from dropdown menu
- added flow mobile to dropdown menu

image

image


bthaile (approved):
Lgtm

PR #860, opened 2023-01-30
catch POST requests and not send error to sentry

bthaile:
catch post requests until requests can be filtered out before hitting service

IT ticket submitted to get a proxy in front of developer portal to filter out spam requests like POSTs.


fotescodev (approved):
LGTM

PR #862, opened 2023-01-30
Supercommands section

fotescodev:
Adding a new section to FLOW CLI Doc called GETTING STARTED

with the Supper-commands doc pointing to https://github.com/onflow/flow-cli/blob/master/docs/super-commands.md


bthaile (approved):
lgtm

PR #893, opened 2023-02-06
better output when manifest isn't found

bthaile:
Frequently log statement indicated a manifest file isn't found, need to track them down. Add better logging.


MrDSGC (approved):
lgtm

PR #907, opened 2023-02-08
fixed overlapping sections

MrDSGC:
-increasing nav bar height from 90 to 110

before
image

after
image


bthaile (approved):
looks good, we'll need to rem to change this after we remove the banner.

PR #946, opened 2023-02-23
update account linking config, cache issue

bthaile:
Update account link config. Not sure why flow-docs.json file can't be found on production and account linking manifest is fallen back on.

### DIFF
--- a/app/data/doc-collections/account-linking.json
+++ b/app/data/doc-collections/account-linking.json
@@ -17,25 +17,25 @@
           {
             "title": "Account Model",
             "tags": ["guide", "account-linking"],
-            "description": "Learn about the account model enables through Account Linking",
-            "href": "/account-linking/guides/account-model"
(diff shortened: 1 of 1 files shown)


app/data/doc-collections/account-linking.json
```diff
@@ -17,25 +17,25 @@
           {
             "title": "Account Model",
             "tags": ["guide", "account-linking"],
-            "description": "Learn about the account model enables through Account Linking",
-            "href": "/account-linking/guides/account-model"
+            "description": "Learn about the account model that is enabled through Account Linking",
+            "href": "guides/account-model"
```
chrisackermann, line 21: Is this correct, since the current path does include `account-linking`?

bthaile: Yes correct, relative paths work and I think more readable. Starting with a slash means absolute path which can also work. Other configs use relative paths so I'd like to be consistent

PR #952, opened 2023-02-24
fix console error with gtag

bthaile:
Seeing this error, I think it's because of reference to gtag method before it's loaded. 

image


justinbarry (approved):
Lgtm

PR #953, opened 2023-02-24
update nft catalog config to force a refresh

bthaile:
Seeing some weridness in production, updating config to force redeploy


justinbarry (approved):
LGTM

Issue #967, opened 2023-03-03
Sidebar Pagination of external links


bthaile:
Issue is that pagination treats all sidebar links as relative links to markdown pages. Pagination should not create links to external links.

Example:
https://developers.flow.com/account-linking/guides/walletless-onboarding

The "Next" link is to an external link that is treated as a relative link and routes the user to broken link.

PR #970, opened 2023-03-03
#967 :bug: Hide external links from pagination

nialexsan:
hide external links from pagination


bthaile (approved):
Lgtm, there are some clean up suggestions but looks good.

PR #1005, opened 2023-04-08
:rocket: Submit Easinteraction tool

LemonNekoGH:
Use this tool, Golang engineers can only call the API function in generated code, no need to write too many duplicate transaction creation code, and Contract engineers can only prepare some functions in contract as API, no need to write too many transaction/query script.
[More information](https://github.com/LemonNekoGH/easinteraction-for-cadence)


nialexsan (approved):

LemonNekoGH:
Thank you very much!

PR #1006, opened 2023-04-14
:rocket: Submit Easinteraction tool (#1005)

nialexsan:
Use this tool, Golang engineers can only call the API function in generated code, no need to write too many duplicate transaction creation code, and Contract engineers can only prepare some functions in contract as API, no need to write too many transaction/query script. [More
information](https://github.com/LemonNekoGH/easinteraction-for-cadence)

### DIFF
--- a/app/data/pages/tools.ts
+++ b/app/data/pages/tools.ts
@@ -44,6 +44,7 @@ import {
   flowviewTool,
   dapperSelfCustodyWallet,
   unitySDK,
+  easiGen,
 } from "../tools"
 
 export const data: ToolsPageProps = {
@@ -64,6 +65,7 @@ export const data: ToolsPageProps = {
     commandLineLinter,
     cdcWebpackPlugin,
     graffleTool,
+    easiGen,
   ],
   wallets: [
     bloctoWallet,
--- a/app/data/tools/index.ts
+++ b/app/data/tools/index.ts
@@ -611,6 +611,22 @@ const unitySDK: Tool = {
     "https://cdn.jsdelivr.net/gh/devicons/devicon/icons/unity/unity-original.svg",
 }
 
+const easiGen: Tool = {
+  repo: {
+    name: "easinteraction-for-cadence",
+    owner: "LemonNekoGH",
+  },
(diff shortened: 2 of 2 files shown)


10thfloor:
Somewhat related, but I asked ChatGPT to generate the a JS class based on the example contract. 

```
pub contract UserProfiles {
    access(self) let usernames: {Address:String}

    pub fun setName(_ acc: AuthAccount, _ name: String) {
        self.usernames[acc.address] = name
    }

    pub fun getName(_ addr: Address): String {
        return self.usernames[addr] ?? ""
    }

    init() {
        self.usernames = {}
    }
}
```

I asked ChatGPT: Given this contract code, can you generate a JS class that uses FCL to call the contract methods?

Here's what it spit out!

```
import * as fcl from "@onflow/fcl";
import * as t from "@onflow/types";

class UserProfiles {
  constructor(contractAddress) {
    this.contractAddress = contractAddress;
  }

  async setName(name) {
    const transaction = `
      import UserProfiles from 0xProfileContractAddress

      transaction(name: String) {
        prepare(account: AuthAccount) {
          UserProfiles.setName(account, name)
        }
      }
    `.replace("0xProfileContractAddress", this.contractAddress);

    const response = await fcl.send([
      fcl.transaction(transaction),
      fcl.args([fcl.arg(name, t.String)]),
      fcl.proposer(fcl.authz),
      fcl.payer(fcl.authz),
      fcl.authorizations([fcl.authz]),
    ]);

    return await fcl.tx(response).onceSealed();
  }

  async getName(addr) {
    const script = `
      import UserProfiles from 0xProfileContractAddress

      pub fun main(addr: Address): String {
        return UserProfiles.getName(addr)
      }
    `.replace("0xProfileContractAddress", this.contractAddress);

    const response = await fcl.send([
      fcl.script(script),
      fcl.args([fcl.arg(addr, t.Address)]),
    ]);

    return await fcl.decode(response);
  }
}

export default UserProfiles;
```

So cool!

nialexsan:
moved this change to https://github.com/onflow/docs/pull/250

